PR #95, opened 2026-09-14
E00-03 Memory stores: episodic (raw artefacts), semantic (facts), working (current episode)

### LINKED ISSUE #3: E00-03 Memory stores: episodic (raw artefacts), semantic (facts), working (current epis
**Epic** E00 Platform and data foundation
**Persona** System · **Tier** T1 · **Priority** P1 · **Size** L

**Story**
Memory stores: episodic (raw artefacts), semantic (facts), working (current episode) with provenance on every fact

**Acceptance**
Every semantic fact links to at least one artefact and a confidence

**Depends on**
none

**Spec** see docs/00-MASTER-BUILD-SPEC.md section 0 for the document that covers this epic.

ashfrnndz21:
Closes #3

**Acceptance:** every semantic fact links to at least one artefact and a confidence. The Session 2 line from TASKS.md is the first test: a Fact without provenance is rejected — by the service (`NoProvenance`) and by a check constraint on the table, so no path that skips the service can slip one in.

**What is here**
- `app/memory/models.py`: `Artifact` (kind, storage key into the profile's region, content type, sha256, captured_at, source channel, region — never bytes), `Event` (kind, occurred_at, optional artefact and episode, a label of at most 80 characters on one line), `Fact` (subject/attribute/value-as-JSON with unit, confidence 0..1 and state extracted / confirmed_by_person / disputed, provenance, valid_from / valid_to, supersedes_id / superseded_at), `Episode` (kind, label, opened_at, closed_at), `Provider` and `Appointment` (the spine). All carry `ProfileScoped`.
- Immutability is enforced by a mapper `before_update` listener: `Artifact` and `Event` take no change; `Fact` takes exactly one, `superseded_at`. Anything else raises `ImmutableRow` (a `Refusal`) at flush.
- `app/memory/episodic.py`: `store_artifact` (refuses a region other than the profile's and a sha256 that is not a digest), `record_event`, `require_artifact` / `require_event`.
- `app/memory/semantic.py`: `assert_fact`, `supersede_fact` (old row stays with `superseded_at`, new row names it via `supersedes_id`; subject and attribute are kept; provenance and unit carry over unless new ones are given, so confirming a reading still links to the photo), `current_facts` (unsuperseded, `valid_from <= at < valid_to`, narrowed by subject / attribute; `at=` asks what held on a day).
- `app/memory/working.py`: `open_episode` (one open episode of a kind at a time), `close_episode`, `open_episodes`, `require_open_episode` (events, facts and appointments attach only to an open episode).
- `app/memory/spine.py`: `add_provider`, `list_providers`, `book_appointment` (writes down a visit a person has arranged — it contacts nobody and books nothing on anyone's behalf), `upcoming_appointments` (planned or confirmed, from now, soonest first).
- Every read and write goes through `app.audit.access`: scope RECORDS for artefacts, events, facts and episodes; VISITS for providers and appointments.
- `migrations/versions/0003_memory_stores.py`, `migrations/env.py` now imports the audit and memory models so autogenerate sees every table.
- Tests: `tests/test_memory_acceptance.py` (the acceptance line, four ways) and `tests/test_memory.py` (artefact region and digest, immutability, short labels, cross-profile provenance refused, supersession history, validity window, episode open / close, the spine, every row profile-scoped and every write in the trail, a caregiver key with [medicines, visits] cannot read facts and the refusal is logged while the spine stays readable).

**Design decisions**
- *Parallel migration.* Revision `0003_memory` revises `0002_audit`, beside the consent story's `0003_consent` which revises the same parent. This deliberately leaves two alembic heads; the operator joins them with a merge revision. `tests/test_migration.py` no longer walks a fixed chain: it loads every revision in the directory, checks there is exactly one root and that every parent exists, then upgrades and downgrades all heads in dependency order (siblings by file name, so the run is stable). A merge revision with a tuple `down_revision` is already handled.
- *Provenance is a table constraint as well as a service refusal.* `ck_fact_has_provenance` (`artifact_id IS NOT NULL OR event_id IS NOT NULL`) and `ck_fact_confidence` (0..1) live in both the model and the migration. The service additionally checks the artefact or event is on this profile (`NoSuchProvenance`), which the database cannot.
- *Provider is profile-scoped, not shared.* Two profiles seeing the same doctor hold two rows, so nothing about one graph can be learned from another. Its `region` is where the provider is (a JB clinic for a Singapore patient is real) and is not guarded; the row itself lives with the profile. An artefact's `region`, by contrast, is where the bytes are and must equal the profile's (`OutOfRegion`).
- *No free text carries content.* `LABEL_LENGTH = 80`, one line, applied to `Event.label`, `Episode.label` and `Appointment.purpose` via `short_label`, which raises `NotALabel`. What was said is in the artefact and only there.
- *Scopes are per store for now.* All facts and events sit under RECORDS. Per-kind scoping (a blood-pressure fact under READINGS, a dose under MEDICINES) can be layered when those modules land without changing callers.
- `Episode` uniqueness (one open per kind) is enforced in the service rather than by a partial unique index, which is dialect-specific.

**Not in scope:** the State model (E00-04), any HTTP endpoints, the recall index. No patient-facing strings.

**Tests and lint**
- `make test`: 42 passed (23 before, 19 new).
- `make lint`: ruff clean, mypy clean (25 source files).
- `make plain-words`: skipped, `app/safety/plain_words.py` is Session 6.


ashfrnndz21:
Operator merge note: CI backend green. The safety reviewer could not run (API credits exhausted); the operator ran the suite locally (42 passed, ruff 0.16.7 and mypy clean) and read app/memory/models.py in full. Provenance is a DB check constraint, artefact/event rows are frozen by a before_update listener and facts change only in superseded_at, labels are capped at 80 chars so no free text can carry content, artefact region must equal the profile's. Migration 0003_memory branches from 0002_audit beside the consent story's 0003_consent; a merge revision follows when consent lands.

PR #121, opened 2026-09-14
E11 Delivery: trigger engine, channels and caps, the escalation ladder, the morning ritual, card grammar, voice twins

ashfrnndz21:
Closes #5, closes #42, closes #43, closes #44, closes #45, closes #46, closes #48
Refs #41 — the family group thread and voice notes *from* Pa are not built here; the issue stays open.

E11 delivery on the modules main already has — E19's WhatsApp door and templates, E21's feed and ranking, E12's roster, the red-flag table, E04's count, E10's routine, E13's button, E05's transcript flags, E01's first-week plan — not beside them.

### #5 E00-05 Trigger engine (rule, pattern, event; per-type channels and caps) — *No persona receives more than the configured cap per day; every trigger logs its rule*
| Already on main | Built here | Deferred |
|---|---|---|
| `run_morning`, `run_visit_card` (E19); the reorder date from E04's count; `open_flags`; review cards; E12's scheduled pushes (nothing delivered them) | `app/delivery/triggers/`: `rules.RULES` — eight types across the three kinds (rule: morning, dose window closed, reorder date reached; pattern: three untapped tablets in seven days; event: red flag, visit tomorrow, paper waiting, family message due), each with scope, channel list, per-type cap and quiet-hours flag; `engine.run_due(profile, at)`, scheduler-free, flags first; a `Delivery` row per attempt naming its rule (sent / capped / quiet / no scope / no channel / skipped), SHARE on the trail when it reached a person; caps per person per type per day, the hold written once; `POST /dev/run-triggers` (dev only) | The scheduler itself: a deployment concern, `*/5 * * * *` over the region's profiles (documented in `engine`, ADR 0005) |

### #45 E11-05 Trigger channels and caps per type — *Reminder, alert and context cards route to the configured channel*
| Already on main | Built here | Deferred |
|---|---|---|
| WhatsApp templates through the one `send` door (E19) | `DeliverySettings` (history rows): per-type channel list and cap, quiet hours, quiet-day skip; `GET/PUT /profiles/{id}/delivery-settings`; an alert's cap is refused (`AlertsAreNeverHeld`); channels app push → WhatsApp template → caregiver, first that can carry it wins, the rest written in `passed_over`; a `PushSender` port (content-free line only: no health data through a phone maker's servers) with `NoDevices` until the app registers devices; six new templates (dose ask, dose check, reorder to the family, the count, papers waiting, family note) | App-push device registration (W1/iOS): until then the app channel falls through by design |

[...]

PR #120, opened 2026-09-14
Security: rows are read under the scope they were written under (artefacts, events, facts)

ashfrnndz21:
Scope was enforced per **table**: `scoped_select(model, context, scope)` checked the scope the caller named and returned every row of that table on the profile. Two tables hold rows of different kinds written under different scopes, so rows leaked across. This makes every artefact and event row carry the scope it was written under, narrows every read by it inside `scoped_select`, narrows every fact reader by the fact's subject scope, and adds a property test that walks every read route for every kind of key.

## The two leaks, with reproduction (both confirmed on main `d67fa23`, both now regression tests)

**1. The family's WhatsApp messages and the questions asked of Nura, readable by a key holding RECORDS only.** WhatsApp coordination messages (E19) are MESSAGE artefacts written under FAMILY; Ask questions (E03) are MESSAGE artefacts written under ASK. Any read of `artifact` under RECORDS returned them.
- Service: Mei sends "who is taking him on Thursday?" (coordination), Pa asks "what was my blood pressure"; the owner cuts a caregiver key narrowed to `records`. `audited_read(Artifact, key, RECORDS)` returned both message artefacts, and `require_artifact` opened either one by id.
- HTTP: once the owner hangs the family message on an episode (`POST /episodes/{e}/attach`), the records-only key saw its artefact reference on `GET /timeline`, `GET /episodes/{e}` and `GET /changes`.
- Test: `test_a_key_holding_the_record_only_cannot_read_the_familys_messages_or_questions`.

**2. `GET /profiles/{id}/facts` with no subject returned medicines and readings to a key holding RECORDS only.** Pa types a reading (`POST /readings`), a medicine is added, a lab photo is confirmed; a caregiver key narrowed to `["records"]` calls `GET /facts` and gets `blood_pressure` (READINGS) and `medication` (MEDICINES) beside `lipid_panel`. It now gets `lipid_panel` only; the subject-named reads for the other two are still refused (403).
- Test: `test_the_whole_record_read_of_the_facts_returns_only_the_parts_the_key_holds`.

## The fix

- **Row scope on artefacts and events.** A `RowScoped` mixin (`app/keys/rows.py`) adds `written_scope` (NOT NULL). `scoped_new` sets it from the scope of the write, and a caller cannot name it (`TypeError`). `scoped_select` adds `written_scope IN context.scopes` in the same expression as the profile filter, so `audited_read` and every other reader through any door returns only rows of the key's scopes.
[...]

PR #119, opened 2026-09-14
E17 Feeling cloud, feeling inference, smart nudges, the proud number, nudge metrics

ashfrnndz21:
Closes #69 #70 #72 #73 #47
Refs #71 (partial: curiosity and context nudges deferred — #71 stays open)

Every acceptance line is met and is a test. After review: a red tap presses the not-feeling-well button server-side and returns its card; one smart nudge a day (docs updated); the check-in time is his `setting.checkin_time`; curiosity and context nudges are deferred.

## Acceptance lines

| Story | Acceptance | Where it is proven |
|---|---|---|
| #69 E17-01 | Words re-rank within a second of a State change; disappears after a tap or Fine today | `test_feelings_cloud.py::test_the_words_re_rank_within_a_second_of_a_state_change`, `::test_the_strip_goes_after_a_tap_or_fine_today_and_comes_back_after_a_change`; CP21 |
| #70 E17-02 | Red flags open not-feeling-well and notify the roster within one minute; every inference names its reason | `test_feelings_inference.py::test_a_red_word_goes_to_the_red_flag_path_first_and_makes_no_note` (E13's path at the tap: flag kept, a notice to each person on his emergency list, the ladder; `opens: not_feeling_well`), `::test_a_yes_that_tells_the_red_variant_apart_takes_the_red_flag_path`, `::test_at_most_two_things_to_tell_and_every_reason_kept`; CP21 |
| #71 E17-03 | Cap enforced per profile per day; every nudge carries why; dismissals feed ranking | `test_nudges.py::test_a_visit_tomorrow_is_anticipated_and_the_rest_wait_one_a_day`, `::test_one_a_day_counts_the_one_already_handed_over`, `::test_a_dismissal_moves_that_kind_down_his_day`, `::test_a_kind_ignored_twice_rests_for_a_week`, `::test_never_at_night`, `::test_none_on_a_day_with_a_red_flag`; CP21 |
| #72 E17-04 | Never decreases; no streak language | `test_proud.py::test_the_number_never_goes_down_whatever_the_days_bring`, `::test_the_me_page_says_it_without_streak_words` |
| #73 E17-05 | Weekly counts per profile | `test_nudge_metrics.py` (taps, Fine-today share, acceptance by kind, ignored streaks; owner and chief only, refusal on his trail), `test_feelings_api.py::test_the_metrics_say_counts_and_nothing_about_his_health`; CP21 |
[...]

PR #115, opened 2026-09-14
W3 Web onboarding: about you, word cloud, read-back, assistant-led records, questions, gaps and unlocks

ashfrnndz21:
Supersedes #112 (closed when its base branch was deleted).

Refs #10 #11 #12 (web half). This is TASKS.md Session 12 (onboarding) on the web, per ADR 0001, now finished over E01's backend (#117). It targets `main`; every merge of `main` into this branch was by exact SHA, the last two `917ce58` (E01, #117) and `12a91db` (#120, per-row scope). Head: `e109f11`.

## This round: finalized on E01 (#117) and #120

- **No mocks.** `web/src/api/mock/`, the client's mock hook, `VITE_API_MOCK` and `make web-mock` are gone. Every route #117 serves is called for real: conditions, settings, the sitting (`/biography`, `/papers`, `/read-back`, `/questions`, `/close`) and the plan (`/plan`, `/plan/later`). The data layer uses #117's shapes (`web/src/api/types.ts`).
- **Kept questions render their `state_id` and `source` line** (`data-state-id` and the provenance slot), and the e2e asserts both against `GET /biography`.
- **Keep this one is one call.** E01 puts a kept question on the next visit's list itself (E05), or keeps it waiting and moves it when a visit is booked (`handed_over_to`). The client's own E05 posts are removed so nothing is added twice. The e2e covers both paths: a kept question with a visit booked lands on its list at once; one kept with none moves there when a visit is booked afterwards.
- **A chief reads in her own language.** E01's writes answer in the language on his settings, so after each write a chief's phone reads the sitting and the week again with `?language=`. Choosing Malay for Pa never turns her screens Malay. E01's script speaks to him (*A few things about you*), so a chief sees the app's own headings in his name (*A few things about Pa*).
- **Tap gaps.** The breakfast gap asks that one question of About you and comes back with the gap closed. Which medicine he is allergic to has no route to write it (see follow-ups), so that card has *Later* only, never a dead button.
- **Migration.** `0019_person_named_by` now revises main's head `0019_row_scope`; one head (`test_migration.py`). The PDPA data-map block is regenerated from the merged models (`--check` clean).
- **#120's route matrix.** `POST /profiles/{id}/consents/sharing/preview` is in `tests/test_row_scope.py`'s `NOT_WALKED`: "renders the words of an agreement the caller sends; returns no rows". No pins were bumped.
[...]

PR #123, opened 2026-09-14
Platform: the suite and migrations on Postgres in CI, a production container, Singapore deploy config, demo mode

ashfrnndz21:
Makes Nura ready to run somewhere other than a laptop, in five parts:
- the whole backend suite and every migration on Postgres, in CI
- a production container, built in CI so it cannot rot
- deploy config for Singapore on Fly.io and on Render
- a demo mode, because none of the outside-world providers is real yet
- a runbook, from creating the hosting account to the home-screen icon

Nothing needs installing on the laptop: Postgres runs only in GitHub Actions, and the image is
built by CI and by the platform. **Nothing is deployed yet**: the owner creates the account
(see `docs/deploy.md`).

**Head:** `3bee0d4`. `main` is merged by exact SHA up to `6e1eb5c` (#121).

**CI on `3bee0d4`: green.**
- `backend`: 10m04s
- `backend-postgres`: 15m30s. It ran migrations up, down to nothing and up again through `0023_delivery`, then `alembic check` on Postgres, then the whole suite.
- `image`: 51s. The build, the release migration, a demo serving as non-root, the refusal to start without demo mode, and the refusal of each demo fixture by name, #121's voice and push included, all passed.
- `web`: 2m54s

The `review` check fails because its API key is out of credits. That is known; leave it.

## 1. Postgres: what broke, and how it was fixed

`NURA_TEST_DATABASE_URL` (`backend/tests/conftest.py`) runs every test on a Postgres. Each
database a test asks for becomes its own schema, created for the test and dropped after it.
Each test engine's connections carry the schema's name as `application_name`.

The new `backend-postgres` job (ubuntu, `services: postgres:16`, health-checked) runs:
- `alembic upgrade heads`, then `downgrade base`, then `upgrade heads` on an empty database,
  through every revision up to `main`'s head, #121's `0023_delivery`
- `alembic check` against #131's registry (`app.models_all`), on the Postgres it has just
  migrated. #131's own test runs the same check on SQLite. Postgres keeps types SQLite does not
  (uuid, timestamptz, json, varchar lengths), so drift SQLite cannot see would show here.
- then the full suite

It took four runs to go green; each row below is what one run found.

| Run | What broke | Why | Fix |
|---|---|---|---|
| 1 | 9 safety tests: `column audit_entry.rowid does not exist` | A test helper (`tests/safety_support.trail`) read the audit trail in SQLite's implicit `rowid` order. The frozen clock makes `at` useless for ordering in tests. | A per-dialect insertion order: `rowid` on SQLite, `ctid` on Postgres. The trail is append-only and each test's schema is new, so `ctid` is insertion order. App code orders by `at` and was never affected. |
| 1 | `test_0005…`: `cannot drop constraint uq_appointment_profile_id_id … other objects depend on it` | The test downgraded 0005 alone with every later revision still applied. Later composite foreign keys depend on 0005's constraints. SQLite's batch rewrite never checked. | Go down through the later revisions first, as `alembic downgrade 0005_memory_review` does. |
| 1 | `test_0012…`: `relation "question" does not exist` | The test applied revisions after 0012 before 0012 itself. It also inserted a red flag with no real person, profile or event, which SQLite allowed because its foreign keys were off. | Apply only the revisions before 0012, and insert real parent rows. |
| 2 | `test_0012`, now reaching its assertion: `[] != '[]'` | A raw `SELECT` of a json column: SQLite returns its text, while asyncpg returns the value decoded. | Compare decoded values. App code reads JSON through typed columns. |
| 2 | The new confirmation race test at teardown: `DROP SCHEMA … lock timeout` | The race itself passed. But the second session, shared across two asyncio tasks, was still inside its transaction when the schema was dropped. | See the race test below. |
| 3 | 16 teardown errors: "left a connection open" | The first version of the new leak check flagged Postgres's own autovacuum workers. It also flagged a test's second schema: `sg` and `my` are two schemas on one server. | The check looks only at client connections of the test's own engine, by `application_name`. |

**Found by reading, before any run:**
- **A platform's database URL** is `postgres://…` or `postgresql://…` with libpq's `sslmode=`.
  `app.settings.database_url_for` gives it the asyncpg driver and `ssl=`, for the app and for
  alembic alike.
- **Postgres engines** get `pool_pre_ping`, because a managed Postgres behind a proxy closes
  idle connections.
- **The SQLite busy-wait** (`BEGIN IMMEDIATE`, lock timeout) is proven not installed on
  Postgres (`tests/test_postgres_engine.py`).

**Checked on Postgres, no change needed** (the whole suite passes there):
- **Columns.**
  - JSON columns: `sa.JSON` becomes `json`, and the server defaults hold.
  - UUIDs are native `uuid`.
  - `enum_column(native_enum=False)` is `varchar(32)` with no CHECK, the same on both.
  - Timezone-aware datetimes are `timestamptz`. Since #130 every datetime column goes through
    `UTCDateTime`, and #130's tests run on Postgres in this job through the `deployment`
    fixture: a visit booked at 09:00+08:00 comes back as 09:00 in Singapore, and a time with no
    offset is refused.
- **Migrations.**
  - `batch_alter_table` runs as plain `ALTER`s.
  - Every named constraint drops cleanly on the way down.
- **Queries.**
  - The one LIKE matches lowercase codes on both.
  - Nothing uses raw `RETURNING`.
- **#120's row-scope matrix passes.**
- **The single-use confirmation, under a real race.** `tests/test_confirm_race.py` is
  Postgres-only. `consume_confirmation`'s own conditional UPDATE
  (`… WHERE id = :id AND consumed_at IS NULL`) runs on two plain connections:
  - The first changes one row and holds it.
  - The second is seen waiting on the row lock in `pg_stat_activity`.
  - After the first commits, the second changes 0 rows. asyncpg reports `UPDATE 0`, which is
    the count the service refuses on.

## 2. The CI `web` failure on `b2465d4`, and its fix

**The failure.** `npx playwright test` failed one spec, `onboarding.spec.ts:110`, #115's
onboarding. It checks that every key the phone keeps in its local store starts with
`session.`, `device.`, `today.`, `proud.`, `takenDays.` or `feed.`. The demo flag was
remembered as `deployment.demo`.

**The cause: this PR, not a flake.** The same spec passes on `main` alone (`d1d057b`), run
locally. The Dockerfile, the Vite base path and the Makefile's dev-server start were not
involved; this PR does not change the Makefile.

**The fix.** The key is `device.demo`, beside `device.language` and `device.density`. It
records what this phone knows about its server, never anything about the person. All six
onboarding specs then pass locally, and `web` has been green in CI since `0703209`.

## 3. Production container

`Dockerfile` at the root has three stages:
- **Web.** Node 20: `npm ci`, then `npm run build`.
- **Dependencies.** Python 3.12-slim: `pip install --no-deps -r backend/requirements.lock`,
  then `pip check`. Every runtime dependency is pinned, and the lock matches the exact pins
  #120 put in `pyproject.toml`.
- **Runtime.** It holds the venv, `app/`, `migrations/`, the fixture answers a demo reads, and
  the web build.

At runtime there is one uvicorn process: the API at `/` and `/api`, the web client at `/app`,
`--proxy-headers`, no access log. It runs as a non-root user (uid 10001) with no dev flags.
Migrations are the platform's release step. `.dockerignore` keeps the context to `web/` and
`backend/`. `tests/test_requirements_lock.py` holds the lock to `pyproject.toml`.

The new `image` job builds the image (never pushed). Then, on an empty `postgres:16`:
- The release step migrates the database.
- A demo container serves `/health/ready`, `/api/deployment` (`demo: true`) and `/app/`. It
  does not run as uid 0, and it refuses a real number with 403.
- The same image started without demo mode exits with `NoCodeSender`.
- The providers the demo deployed, built inside the image and put to the central check with production settings, are each refused by name, including #121's voice and push.

The new endpoints:
- `GET /health` says the process is up.
- `GET /health/ready` also asks the database, and answers 503 when it does not answer.
- `GET /deployment` gives the region and whether it is a demo.

**Consult recordings (#128).** A recording is up to 48 MiB. It is sent on Stop as its own bytes (`audio/webm`) in one request.
- The app refuses anything larger (`MAX_CONSULT_BYTES`), and nothing in the image sets a lower limit: uvicorn has none.
- #128's route refuses a declared `Content-Length` over 48 MiB before reading a byte. A chunked upload with no declared length is read in full before its size is checked. That is #128's code, left as it is here; open question 14.
- The demo guard reads only JSON bodies, so it never buffers a recording.
- The bucket store gives each request two minutes, up from 30 seconds.
- The demo config carries `NURA_SPEAKER_FIXTURES`.
- Whether each platform's own request-size limit allows 48 MiB is open question 13.

## 4. Fixture refusals

Every fixture class is now marked `@fixture`, a registry in `app/fixtures.py` that adds no
attribute to the class, so the calendar port's shape test is unchanged. `create_app` refuses
any of them unless the process is a declared dev run or a declared demo.
`tests/test_fixture_refusals.py` walks `app/` for every class named `Fixture…` and fails if
one is unmarked.

| Provider | Class | Refused outside a dev run before this PR? | Now |
|---|---|---|---|
| Login code sender | `LoggingCodeSender` | yes | yes. A demo uses `DemoCodeSender`, which is refused outside a demo. |
| WhatsApp | `FixtureProvider` | yes | yes (dev run or demo) |
| Drug registry | `FixtureRegistry` | **no**: `NURA_DRUG_REGISTRY` defaults to `fixture` | yes |
| Reference ranges | `FixtureRanges` | **no**: defaults to `fixture` | yes |
| Extractor (OCR / vision) | `FixtureExtractor` | **no** | yes |
| Summariser | `FixtureSummariser` | **no** | yes |
| Transcriber (speech, voice notes) | `FixtureTranscriber` | **no** | yes |
| Speaker separation (#128) | `FixtureSeparator` | new with #128 | yes: marked when merged in here |
| Feed search and compress | `FixtureSearcher`, `FixtureCompressor` | **no** | yes |
| Calendar | `FixtureCalendar` (tests only; a deployment reads `.ics` uploads) | n/a | yes, if ever wired |
| Ask retriever | `FixtureRetriever` (tests only) | n/a | yes, if ever wired |
| Object store | `LocalObjectStore` | **no** | yes. A bucket (`S3ObjectStore`, new: SigV4 over httpx, no SDK, checked against AWS's published signing examples) is required outside a dev run or demo. |
| Voice, a card said aloud (#121) | `FixtureVoice` | dev run only | dev run or demo. `voice_for` refused a demo before, which would have stopped a demo starting. |
| App push (#121) | `FixturePush` (dev run), `NoDevices` (everywhere else) | `NoDevices` ran anywhere | yes: a demo runs on `NoDevices` and reaches nobody; production refuses until a real sender exists |
| Frozen clock, `/dev/…` routes, `?at=` | — | yes (dev run only) | unchanged, and off in a demo |

## 5. Demo mode (`NURA_DEMO_MODE=1`, `docs/adr/0008-demo-mode.md`, status: accepted)

- **Fixtures.** It is the only way to run on the fixtures anywhere but a laptop.
- **The banner.** Every web screen starts with "Demo — not for real health information",
  followed by "This is a demo. Do not put real health information in it. Everything here is
  wiped each night." It is in English, Malay and Chinese, and passes plain-words and
  `make language`. It is remembered for offline. The printable emergency card prints it first,
  and the API's page title says demo.
- **Test numbers only.** Only `+65 0xxx xxxx` and `+60 0…` are accepted; neither country gives
  a subscriber number beginning with 0. This is enforced twice:
  - the demo code sender refuses any other number before a challenge is written
  - `DemoNumbersOnly` refuses any JSON body carrying a real-looking number, on any route (403
    `NotInTheDemo`)

  Email sign-in is refused.
- **Sign-in.** Nobody is sent a code. A test number signs in with the operator's secret code,
  `NURA_DEMO_LOGIN_CODE`, and the code is never logged.
  Since #132 the sign-in message is written in the language the screen sends (en, ms, zh). On a
  demo it goes nowhere: the demo sender takes it and neither sends nor logs it, and a test
  number signs in with the operator's code in each of the three (`tests/test_demo_mode.py`).
- **The nightly wipe.** At 03:00 on the region's wall clock every row is truncated, keeping the
  schema, and the local object directory is emptied. The check runs before serving and every
  five minutes after. No state is kept.
- **Not a dev run.** Demo mode and a dev run are exclusive; the frozen clock and the dev routes
  stay off.

**Checkpoints walked against a local demo-mode server** (fixtures, test numbers, the
operator's code):
- 2, 3, 4, 5, 6, 14, 16, 17, 18 and 21 pass. The walkers take `NURA_DEMO_LOGIN_CODE`, including
  #119's cp21.
- 7 stops, as designed, at its `?at=` feed step (`NotOnADevRun`).
- 8, 9, 13 and 15 also need a dev run.

## 6. Deploy config (Singapore only) and runbook

**Render is the recommended platform for the demo.** Nothing is installed on the laptop, and Render's Postgres is in Singapore. Artefact bytes stay on the instance's disk, for the demo only. Fly.io is the alternative.

- **`fly.toml`**:
  - app `nura-sg`, `primary_region = "sin"`, internal port 8000
  - `release_command = "alembic upgrade heads"`
  - health check on `/health/ready`
  - no volume: bytes go to a bucket through `NURA_OBJECT_BUCKET_*`
  - one machine kept running for the nightly wipe
- **`render.yaml`**:
  - a Docker web service in `singapore`, plan `starter` (the pre-deploy migration needs a paid
    instance)
  - a managed Postgres 16 in `singapore` with an empty IP allow-list
  - the demo code as a `sync: false` secret
- **`docs/deploy.md`** covers (Render first, as §4; Fly.io as §5):
  - what is deployed, and what cannot be real yet
  - every setting
  - the first deploy on Fly and on Render, and rollback
  - the checkpoints over https: `NURA_BASE_URL=https://… NURA_DEMO_LOGIN_CODE=…`
  - the phone and the home screen
  - what must be true before real health information
- **`docs/checkpoints.md`**: row 19 is now "Nura on your phone over https (demo)", with a how-to
  pointing at `docs/deploy.md`. TestFlight is the unnumbered last row, "later — needs an Apple
  developer account". Row 19 names Render as the recommended host. The PDPA map's tabletop and DPO lines now say "before the first family on
  real health information".

### Settings

| Variable | Secret | Demo | Real deployment |
|---|---|---|---|
| `NURA_REGION` | no | `SG` | `SG` |
| `NURA_DATABASE_URL` | **yes** | the platform's Postgres (direct, not pooled) | same |
| `NURA_DEMO_MODE` | no | `1` | **absent** |
| `NURA_DEMO_LOGIN_CODE` | **yes** | six digits | **absent** (refused) |
| `NURA_OBJECT_BUCKET_URL` / `_REGION` | no | Fly: Tigris URL and `auto` | a Singapore bucket, `ap-southeast-1` |
| `NURA_OBJECT_ACCESS_KEY_ID` / `_SECRET_ACCESS_KEY` | **yes** | from the bucket | from the bucket |
| `NURA_OBJECT_STORE` | no | Render demo: `/tmp/nura-objects` | **absent** (refused) |
| `NURA_PAPER_/VISIT_/VOICE_/FEED_/WHATSAPP_FIXTURES` | no | `tests/fixtures/…` | absent once real adapters exist |
| `NURA_WHATSAPP_DEV_SECRET` | **yes** | random | **absent** |
| `NURA_WHATSAPP_PROVIDER`, `_NUMBER`, `NURA_DRUG_REGISTRY`, `NURA_REFERENCE_RANGES` | no | unset (fixture) | the real names, once built |
| `NURA_REVIEW_STAFF_TOKENS` (#125) | **yes** | unset | the pharmacist review queue's staff tokens |
| `NURA_WEB_DIST` | no | set by the image | same |
| `PORT` | no | 8000 (Fly); Render sets its own | same |
| `NURA_DEV_CODE_SENDER`, `NURA_FROZEN_CLOCK` | — | **must be absent** | **must be absent** |

## Checks

Run locally on `3bee0d4`, which is `main` at `6e1eb5c` (#121) merged:

- `make test` (SQLite): 1401 passed, 1 skipped (the Postgres-only race test). The 45 warnings
  are Python 3.12 deprecating SQLite's default datetime adapter, raised inside SQLAlchemy;
  none comes from this PR's code.
- `make lint`: ruff clean, `mypy app` clean (220 files).
- `make language` (#125): 0 failures across 4,653 strings in 24 catalogues. None of its notes
  is on this PR's lines, and the step stays in the `backend` job.
- Web: Vitest 172 passed (20 files), `npm run plain-words` 0 failures. The typecheck printed no
  errors, and CI's `web` job runs it again in `npm run build`.
- `make plain-words`:

```
plain-words: 4548 strings in 26 files, 0 failures, 58 notes, 1 exempt as history
```

  Every new patient line is in en, ms and zh: the demo banner, its lines, and the
  `NotInTheDemo` refusal. The notes (lines over ten words) are all on lines that were there
  before this PR.
- Checkpoints walked against a local demo-mode server: 2, 3, 4, 5, 6, 14, 16, 17, 18 and 21
  pass. 2–6 walked again after #132, and 2 again after #121, with a demo booting on its voice and push: all five pass, and a test number starting sign-in with `language: "ms"` gets 202.

## Decided in review (the operator, on `8eb1a6d`)

- **Hosting.** Render is the recommended demo platform, and `docs/deploy.md` puts it first. Fly.io
  is the alternative.
- **ADR 0008 is accepted.** The shared demo code is for demo mode only, with test numbers no phone
  can have, the banner everywhere, a nightly wipe, and no real data.
- **uvicorn is pinned to 0.35.0** in `pyproject.toml`, the version the image runs.
- **ADR numbering is fine.** 0005 arrives with #121, and 0006 and 0007 are on `main`.
- **The `review` check** fails because its API key is out of credits. That is known; leave it.
- **Required checks.** `backend-postgres` and `image` become required checks after the merge.
- **The chunked-upload read in #128's recording route** is filed as its own issue.

## Before real data (ADR 0008, `docs/deploy.md` §8)

- The bucket is Singapore-only. Tigris and the instance's disk are for the demo only.
- The platform's request logs are checked: where they are kept and for how long. URLs carry
  profile ids, never names or numbers.
- The platform accepts a 48 MiB request, the largest recording of a visit.
- Real providers in the region: SMS or email, a WhatsApp business solution provider, drug data,
  reference ranges, OCR and vision, speech, a summariser, and the feed's fetcher and model.
- The data protection officer is named, and the breach tabletop has been run.

Then, and only then, `NURA_DEMO_MODE` comes out.

## Still open

- **Fly Postgres in Singapore**, only if Fly is ever chosen: is Fly Managed Postgres offered in
  `sin`? If not, the runbook says to stop rather than create it elsewhere.


ashfrnndz21:
Operator merge note: CI backend, backend-postgres, image and web green on 3bee0d4 (main through 6e1eb5c; main since moved by docs only); the GitHub reviewer is down, so the operator reviewed in its place over three rounds. risk:high (deploy, fixtures refused in production). Operator decisions: Render is the recommended demo platform (nothing to install), Fly the alternative; Tigris or instance disk for the demo only, a Singapore-only bucket before real data; ADR 0008 accepted (shared demo code on non-dialable test numbers, banner everywhere, nightly wipe, no real data); uvicorn pinned 0.35.0. Accepted: the suite and migrations (up/down/up, alembic check) on Postgres in CI; every fixture provider marked and refused outside a dev run or demo, E11's voice and push included, proven inside the built image; a non-root pinned container; postgres:// URLs converted and connections pre-pinged; demo sign-in never sends or logs a code. Builder's local gate 1401 passed. Next: backend-postgres and image become required checks.

PR #125, opened 2026-09-14
E22 Same words every time: glossary and translation memory, voice scripts, the pharmacist review queue

ashfrnndz21:
Closes #87 #88 #89

## Acceptance lines

- **E22-02 (#87)** — *Term consistency enforced across cards, voice and WhatsApp.* `make language` (wired into CI's backend job after plain-words) holds every patient string in every catalogue — the feed cards, WhatsApp replies and templates, the web client — to one Malay and one Chinese line per English line and to the glossary's words: **0 failures**. The voice is the cards' own lines said (E22-03), so it cannot drift from them.
- **E22-03 (#88)** — *Every patient card has audio within a minute of creation.* Every card now carries its **voice script**, computed from the same verified lines at the moment it is served, with a digest to key pre-rendered audio. The audio itself is E11's `Voice` port (PR #121, not on main): the seam is documented and the acceptance is met in full when #121 lands (see *Follow-ups*).
- **E22-04 (#89)** — *Nothing from a new source reaches a patient before review.* A source proposed to the pharmacist's queue is pending and not allowlisted; a self-search naming it is refused (`SourceNotAllowlisted`) until staff approve it. Tested over HTTP and walked in CP23.

## What is here

**#87 Glossary and translation memory** (`backend/app/language/`)
- `glossary.py` — docs/plain-words.md as a table: §2 (red word → plain phrase), rule 13 (his word → variants never said), and a new **§6 "The same words in Malay and Chinese"**: 27 rows, English | Malay | Chinese | never-said variants per language. The Malay and Chinese are the words the catalogues already shipped, chosen where they disagreed — no model, no translation service.
- `memory.py` — every tagged patient string (the same files and `@patient` tags as the plain-words verifier) in one table keyed `catalogue:key:language`, e.g. `backend/app/delivery/strings:LINES.flag_family.1:zh`, `web/src/strings:refusals.OutOfScope:ms`. The checker: `twins` (every en line has ms and zh twins, card by card where Chinese splits a line for rule 10), `slots` (twins fill the same slots), `phrase` (one English line is one ms and one zh line wherever it is filed), `term` (§6 words used), `never` (§6/rule 13 variants absent — "your record", 医院信). Findings in words that change only as a new consent version or a re-approved WhatsApp template are **notes with their follow-up**, never failures. `python3 -m app.language.memory --table` prints the memory.
- `make language`, CI step after `make plain-words`.

**#88 Voice scripts** (`backend/app/language/voice_script.py`)
[...]

PR #130, opened 2026-09-14
Times keep their offset: store every datetime as UTC

ashfrnndz21:
## The bug

`app/db.py` mapped every `Mapped[datetime]` to `DateTime(timezone=True)`. Postgres keeps the offset; SQLite (laptop and CI) drops it: SQLAlchemy's SQLite type writes the wall-clock fields and throws the offset away, and `as_utc` read the bare value back as UTC. A visit booked at `09:00+08:00` was stored as `09:00`, came back as `09:00 UTC`, and was said to the patient as 5 in the afternoon. Every datetime column was affected (appointments, events, facts' validity, delivery times, audit timestamps).

## The fix

- **`app.db.UTCDateTime`**: one `TypeDecorator` (impl `DateTime(timezone=True)`, `cache_ok = True`), pointed at by the type map's `datetime` entry, so every `Mapped[datetime]` column uses it with no model changes.
  - bind: an aware datetime is converted to UTC; a **naive datetime is refused** with `NaiveDatetime` (never assumed UTC);
  - result: always aware UTC (`timezone.utc` attached to a bare SQLite value, which is UTC by construction).
- **No explicit `DateTime(...)` columns** exist in `app/` (the only one was the type map). Migrations spell `sa.DateTime(timezone=True)`; the SQL type is unchanged, so no migration. `alembic check` on a fresh DB at heads shows no type change (its only ops are pre-existing: tables whose models `migrations/env.py` doesn't import — `emergency_card`, `notice`, `what_to_do_card`, `review_item`).
- **Client input datetimes are `AwareDatetime`** (body fields and query params in `app/channels/api/`), so a client sending a time with no offset gets a 422 at the door instead of a 500 from the refusal. `dev_clock` already refused naive input and is untouched. The web client sends `toISOString()` (always `Z`) and the checkpoint scripts send aware times, so nothing current is affected.
- **`scripts.data_map`** names a decorated type by the type it stores as, so the PDPA data map still says `DateTime` and the doc is unchanged.
- **ADR**: `docs/adr/0009-times-are-stored-as-utc.md`.
- Covers #125's `review_item.created_at` / `decided_at` (merged main at `7591ee3`): both come through the type map, written with `utcnow()`.

## Tests (`backend/tests/test_utc_datetime.py`)

1. Book a visit at `09:00+08:00` through `POST /profiles/{id}/appointments`, read it back through `GET /profiles/{id}/appointments`: same instant (`01:00Z`); the brief says "9 in the morning", nothing "in the afternoon".
2. A naive `scheduled_at` sent to the API is a 422.
3. Round trip through a `UTCDateTime` column for `+08:00`, UTC and `-05:00`: the same instant comes back, as UTC, and an SQL comparison at another offset finds it.
4. Writing a naive datetime raises (`StatementError` wrapping `NaiveDatetime`).
5. Every datetime column on the app's metadata is `UTCDateTime` (none a bare `DateTime`).

With main's column type patched back in, tests 1 and 5 fail (`09:00+00:00` came back where `01:00+00:00` was booked); with the fix they pass.

## Naive writes found

**None.** With the refusal in place the full suite (1136 tests before merging main, 1222 after) surfaced no naive write in app, tests or fixtures: ruff's `DTZ` rules (in `backend/pyproject.toml`) already ban naive `datetime()`, `now()`, `utcnow()`, `today()` and `strptime`, and every write goes through `app.clock`. The parse paths ruff can't see were checked by hand: `ingestion/readings.py` attaches the patient's zone to a screen time with no offset; `feed/rank.py` refuses a naive cursor; `timeline.py` / `feelings/record.py` read back with `as_utc`; `family/roster.py`'s `replace(tzinfo=None)` is on `time` values, not datetimes.

## Checks

- `python3 -m pytest -q -p no:cacheprovider`: 1222 passed, 0 failed (after merging main)
- `make lint`, `make plain-words`, `make language`: green
- Fresh DB (`make reset-db`) on port 8119: `make checkpoint N=7` (18 ✓), `N=16` (15 ✓), `N=17` (17 ✓); there is no checkpoint 22.

Existing `dev.db` files written at a non-UTC offset keep their wrong instants; `make reset-db` starts over. Postgres rows were always right.


ashfrnndz21:
Operator merge note: CI backend and web green on 18e1731; the GitHub reviewer is down, so the operator reviewed in its place. Accepted: one UTCDateTime type in the declarative map (aware to UTC on the way in, naive refused with NaiveDatetime, aware UTC on the way out), no schema change (alembic check shows no type diff), API input datetimes as AwareDatetime so a naive client time is a 422 not a 500, ADR 0009. The booking and column-type tests fail with main's old type swapped back in, so they catch the bug. Builder ran make test (1222), lint, plain-words, language, checkpoints 7, 16, 17 on a fresh DB. Existing local dev.db files keep wrong instants until make reset-db. Follow-up: migrations/env.py doesn't import every model, so autogenerate would drop tables (a separate PR).

PR #128, opened 2026-09-14
Visit day: logistics card, in-visit recording with consent, speaker separation, clips

ashfrnndz21:
Closes #35, closes #36, closes #17, closes #25

The visit day, end to end: the logistics card for the next visit (E05-03), one tap to record it with the consent prompt first (E05-04), the recording kept as a consult, transcribed, split by speaker with timestamps (E02-05), and the clip a recall citation points at, played on a tap (the remaining half of E03-05). Web-first (ADR 0001); no new dependency; no live speech provider — ports with fixtures.

## Acceptance

| Story | Acceptance | Where it is held |
|---|---|---|
| #35 E05-03 | Driver pulled from roster; card sent T-1 and T-0 | `test_the_logistics_card_is_composed_from_the_record_and_the_roster_waits_for_a_yes`, `test_the_chiefs_yes_gives_the_drive_and_the_card_names_the_driver`, `test_the_logistics_card_comes_the_day_before_and_on_the_day`; e2e `visit.spec.ts` (T-0 card, the yes on the Visit screen); CP22 (T-1 card) |
| #36 E05-04 | One tap to start with consent prompt; ends into post-visit | e2e `visit.spec.ts` test 1 (Start → the RECORDING words → notice → yes → Stop → the post-visit card); `test_nothing_records_without_the_recording_consent` |
| #17 E02-05 | Consent captured before recording; transcript searchable; clip playable at a timestamp | `test_consent_then_notice_then_upload_keeps_a_consult_and_ends_in_the_post_visit_card`, `test_an_answer_cites_the_clip_and_the_clip_plays_only_under_the_recordings_scope`; e2e clip tap (`#t=19.8,28.9`) |
| #25 E03-05 | Answers cite the artefact; consult answers play the clip at the timestamp | recall cites `artifact` with `start_s`/`end_s` and carries `clip`; Ask's *Hear what Dr Tan said* plays it (`ask.test.ts`, the shared `ClipPlayer`); e2e asks "what did Dr Tan say about the water pill" → 19.8–28.9 s |

## What is here

**Backend**
[...]

PR #131, opened 2026-09-14
Alembic sees the whole schema: one model registry, checked in CI

ashfrnndz21:
## The bug

`backend/migrations/env.py` imported a hand list of model modules, and the list had fallen behind. Two modules that declare tables were missing:

- `app.safety.models`: `emergency_card`, `notice`, `what_to_do_card`
- `app.language.models`: `review_item` (#125)

Alembic compared the migrations against an incomplete metadata, so `alembic check` on a freshly migrated database proposed **dropping those four tables** (and their indexes). The next autogenerated migration would have been written without them. Found while checking #130.

## The fix

- **`app/models_all.py`**: one registry that imports every module declaring a table (24 modules) and exposes `metadata`. `env.py` now imports it and compares against `models_all.metadata` instead of keeping its own list.
- **No migration.** With the registry complete, `alembic check` on an empty database upgraded to heads reports *No new upgrade operations detected*, so no model differs from its migration. That still holds after merging #128 (`0022_visit_day`), whose tables sit in modules the registry already imports.

## Tests (`backend/tests/test_models_registry.py`)

Both run in a fresh interpreter: the test process has already imported every model (and test-only tables) through the fixtures, which would hide a missing import.

1. **The registry is complete.** Every class with a `__tablename__` under `app/` is found from the source (AST, no imports), and each module that declares one must be loaded by importing `app.models_all` alone. A new table module left out fails with `add these to app/models_all.py: [...]`.
2. **The migrations build exactly the models.** `alembic upgrade heads`, then `alembic check`, against an empty SQLite database (inherited `NURA_*` stripped). Any diff fails, whether it comes from a missed import or from a model and migration that disagree.

With `import app.safety.models` removed from the registry, both tests fail (the second with `remove_index … notice …`); with it back, both pass.

## Checks

- `python3 -m pytest -q -p no:cacheprovider`: 1242 passed, 0 failed (after merging main at `82106d2`, #128)
- `make lint`, `make plain-words`, `make language`: green
- Fresh DB (`make reset-db`, migrated through the new `env.py`) on port 8119: `make checkpoint N=7` (18 ✓), `N=16` (15 ✓), `N=17` (17 ✓)

Scope is `env.py`, the registry and the test, so that #121 and #128 can add migrations alongside. `scripts/data_map.py` keeps its own complete import list; it could move to `app.models_all` in a follow-up.


ashfrnndz21:
Operator merge note: CI backend and web green on 5e0f767; the GitHub reviewer is down, so the operator reviewed in its place. Accepted: app/models_all.py imports every module that declares a table, and migrations/env.py compares against it (it was missing app.safety.models and app.language.models, so autogenerate proposed dropping emergency_card, notice, what_to_do_card and review_item). Two subprocess tests: every class with a table name loads from the registry alone, and upgrade heads plus alembic check on an empty database shows no diff. Both fail with one import removed. No migration needed. Builder ran the suite (1242), lint, plain-words, language, checkpoints 7, 16, 17. Follow-up: scripts/data_map.py can use the registry.

Issue #129, opened 2026-09-14
Consult recording: upload in chunks so a long visit survives a dropped connection


ashfrnndz21:
Visit day (#128) records the consult on the phone and uploads once, on Stop (up to 48 MB or 90 minutes). A dropped connection at Stop loses the upload, and a hidden page loses the whole recording. Upload in chunks as it records (resumable, each chunk under the recording's consent and region), assemble server-side, and keep the rule that a 'no' or leaving before the doctor answers keeps nothing: nothing is stored until the doctor's yes. Tests: a dropped connection mid-visit resumes; a no discards chunks already sent.

PR #135, opened 2026-09-14
Uploads are read against their cap as they arrive, never after (#133)

### LINKED ISSUE #133: Recording upload: enforce the size cap while streaming, not after reading into memory
Found by the platform review of #123: the consult recording upload route from #128 reads an upload that does not declare its length (chunked transfer, no Content-Length) fully into memory before checking the 48 MiB cap. A client can make the server hold an arbitrarily large body.

Fix: stream the request body to the object store (or a temp file in region) with a running byte count, and abort with the route's plain-words 413 as soon as the cap is passed; nothing is kept from a refused upload. Apply the same to every upload route (capture photos, PDF import, voice notes) through one helper. Tests: a chunked body over the cap is refused without being held in memory (assert on bytes read), and a body under the cap still stores exactly.

Related: #129 (chunked, resumable consult upload) should build on the same helper.

ashfrnndz21:
Closes #133

## The bug

An upload whose body does not declare its length (chunked, no `Content-Length`) was read whole before its size was checked. On the consult recording, `await request.body()` ran before the 48 MiB check; on the JSON uploads, FastAPI parses the whole body before the route or any dependency runs, so the schemas' `max_length` bounds came after the bytes were already held. A client could make the server hold as much as it liked.

## The fix

**One helper, `app/channels/api/uploads.py` `read_capped`.** It is used by every upload.
- It refuses at once when the body declares more than the cap, before a byte is read.
- Otherwise it counts each chunk as it arrives and refuses with the route's own 413 the moment the count passes the cap.
- What it has read waits in a spooled temporary file: in memory up to 1 MiB, then on the disk of the process, which runs in the region it serves.
- A refused upload's spool is closed and gone, so nothing of it is kept.

**Consult recording** (`POST …/appointments/{id}/recording`): the route reads its raw body with `read_capped`, inside the same guard that writes the refusal on the trail (`ConsultTooLong`, as before).

**JSON uploads:** `UploadCaps` is a thin ASGI layer in front of the app (under `/` and `/api`). It reads the body with the same helper before FastAPI parses it, then hands the route the body it read. Each route's cap is the base64 size of its limit plus 64 KiB for the other fields, and past it the layer answers with the route's refusal:

| Route | Refusal |
|---|---|
| `/photos`, `/readings/photo`, `/biography/papers` | `PhotoTooLarge` |
| `/imports` | `PdfTooLarge` |
| `/events/{id}/notes` | `NoteTooLarge` |
| `/documents` | `DocumentTooLarge` (new) |
| `/not-feeling-well`, `/symptoms` | `VoiceNoteTooLong` |
| `/appointments/{id}/transcript` | `TranscriptTooLarge` |
| `/connectors/{id}/scan` | `CalendarTooLarge` (new) |

- The two new refusals subclass `NotADocument` and `NotACalendar` and are mapped to 413. The services' own checks are unchanged.
- A body just over the decoded limit still reaches the route, which refuses it on the trail as before. Only a body past the wire cap is refused at the door, before any key is resolved, so that refusal is not on a trail.
- `/biography/papers` can carry a PDF, but its schema already bounds `data` at the photo size, so the photo cap matches what it accepts.

## Tests (`backend/tests/test_upload_cap.py`)

Each test sends a chunked body with no length and counts the bytes the server took from it.

- **A chunked photo over the cap**, under `/` and `/api`: refused 413 `PhotoTooLarge`. The server read at most one chunk past the cap, never the four times on offer, and no artefact was kept.
- **An import declared over the cap:** refused 413 `PdfTooLarge` with **0 bytes read**.
- **A chunked recording over the cap:** refused 413 `ConsultTooLong`. It read at most one chunk past the cap, kept no recording or voice artefact, and put the refusal on the trail.
- **A chunked photo (1000-byte pieces) and a chunked recording (7-byte pieces) under the cap:** the object store holds exactly the bytes sent.
- **Every route** whose JSON body carries bytes (`data`, `audio` or `ics`, including a union body) has a cap under both mounts that answers 413. There are 10 such routes.

## Checks

- `python3 -m pytest -q -p no:cacheprovider` on SQLite: 1408 passed, 1 skipped (after merging main at `6a0acc6`). The `backend-postgres` CI job runs the same suite on Postgres.

- `make lint`, `make plain-words`, `make language`: green.
- Fresh DB (`make reset-db`) on port 8119, after merging main at `6a0acc6` (#123): `make checkpoint N=5` (18 ✓, photos and review cards), `N=18` (19 ✓, handwriting, PDFs, notes, device screens), `N=22` (22 ✓, the consult recording).

## Not in this PR

- The WhatsApp webhook (`POST /whatsapp/webhook`) also reads its body whole before checking the signature. It isn't an upload route, but the same cap would suit it; that's a follow-up.
- #129 (chunked, resumable consult upload) can build on `read_capped`.


ashfrnndz21:
Operator merge note: CI backend, backend-postgres, image and web green on 89b78c5; the GitHub reviewer is down, so the operator reviewed in its place. Accepted: one read_capped helper streaming every upload to a temp file with a running count, refusing on a declared length over the cap and with the route's plain-words 413 as soon as the count passes it, nothing kept; the consult recording through it with its refusal on the trail; a thin layer in front of the nine JSON upload routes. Accepted behaviour change: a body far over its cap is refused before the key is resolved and so is not on his trail — a flood of junk must not write rows to anyone's trail; a body just over still reaches the route and is recorded as before. Builder ran the suite (1408, 1 Postgres-only skip), lint, plain-words, language, checkpoints 5, 18, 22. Follow-up: the WhatsApp webhook's body read (new issue).

PR #138, opened 2026-09-14
W7 Web day: brief, questions, visit summary, not feeling well, symptoms, feelings, nudges, clips

ashfrnndz21:
The patient's day in the web client, like for like with the iOS plan (ADR 0001): Today, Visit and Me, and the safety paths. Every line he reads is the backend's, in the backend's order; the client composes no sentence. Checkpoint 27 is the walk (`docs/checkpoints.md`), and ADR 0012 records the offline cards.

The backend changes are small, and each adds something that was missing:
- `GET /profiles/{id}/not-feeling-well/offline` returns the two fixed cards the phone keeps for when it has no network.
- `GET /profiles/{id}/nudges?day=` returns the day's handed-over nudges and what this person did with each. A push carries no words, so this is where the app reads them.
- The feed's memo card now carries `cite.clips`.
- Every follow-up answer carries `red`.
- `POST /symptoms` returns the urgent card when it raises a flag.
- Delivery skips a nudge he already answered in the app.

Both new read routes are in `tests/test_row_scope.py`. There are no new models and no migrations (`alembic check`: no new upgrade operations).

## Stories: acceptance, what proves it, what is not met

| Story | Acceptance | On the web | Proved by | Not met |
|---|---|---|---|---|
| E05-01 pre-visit brief (#33) | Generated at T-3 days; editable; one page | Visit, then *Read before your visit*: the whole brief in the backend's order (purpose, what changed, questions, what to bring), with its boundary last. | `day.spec.ts` › the visit: the whole brief in its order…; `day.test.ts` › shows the whole brief in its order | Unchanged backend gaps: there is no T-3 trigger (the brief is built on read or inside the week); the brief is never edited, only its questions are (E05-02); nothing enforces one page. |
| E05-02 questions card (#34) | Questions carry their source; Dad's card is one screen | His card as the backend wrote it. He adds a question in his own words and it is kept only on a yes to exactly those words; he takes one off on a yes to exactly that. Words that are not plain get the backend's refusal. | `day.spec.ts` › the visit: …his questions added and taken off on his yes; `day.test.ts` › marks each question line… | — |
[...]

PR #141, opened 2026-09-14
Web Push for the home-screen app: reminders on this phone, no health words in a push

ashfrnndz21:
## What this does

The iOS app's push is replaced by **Web Push** for the home-screen app (ADR 0001: iOS 16.4 and later push to an installed PWA). #121's `WebPush`-shaped hole in the push port is filled with a real sender. #123's deployment rules now hold for it.

## Backend

**The sender, `app/delivery/push.py` `WebPush`.** It sits behind #121's port, beside `FixturePush` and `NoDevices`, and is not a fixture.
- The payload is encrypted to the browser's keys with Mozilla's `http_ece` (RFC 8291 / 8188, aes128gcm).
- Every request is signed with Mozilla's `py_vapid` (RFC 8292). The token's `aud` is the push service's origin, and it lasts 12 hours (RFC 8292's limit is 24).
- Requests go out with the `httpx` already in the project. `requests` is not added.

**The payload holds no health words.** It is exactly `{"text": "Nura has something for you.", "id": "<uuid>"}`:
- The text is #121's `PUSH_LINE` in his language.
- The id is, in order: the card on his feed when the trigger names one; for a nudge, the nudge's own id, which is what #138's `GET /profiles/{id}/nudges?day=` returns; else the delivery row's own id.
- The app opens at `/app/?card=<id>` and reads what the push is about from its region. For a nudge it lands on Today, where #138 shows the day's nudge from that route.

**A nudge he already answered in the app is never pushed.** #138's delivery skips it before any channel is tried (a SKIPPED row, "answered in the app"), so no push goes out for it.

**Sending follows #121's rules.** Pushes go under its caps and quiet hours, in each trigger's channel order. A person with no live subscription falls through to WhatsApp, as before ("app_push: no device").

**When a push service answers 404 or 410,** the subscription is dropped (`gone_at`) and that send falls back to WhatsApp ("app_push: gone").

**The port change (approved).** `reachable` and `push` take the delivery run's session, its key context, and a payload id (`ref`). The docstring says why: a person's devices are profile-scoped rows, read like any other row of his. Delivery is the only caller. The fixture and `NoDevices` ignore the new arguments.

**Subscriptions: the `push_subscription` table, migration `0024_push_subscription`.**
- One row per profile, person, login session and browser, under `Scope.PROFILE` (the face of the graph every key opens).
- Subscribing the same browser again replaces the row before it.
- Stopping reminders revokes it.
- A push goes only to a subscription whose login session still stands, so a phone signed out gets nothing.
- The rows are frozen apart from `revoked_at` and `gone_at`.

**The door:**
- `POST /profiles/{id}/push-subscriptions` returns 201 with the subscription's id. Anything that isn't an https endpoint with a P-256 point and a 16-byte secret is `NotAPushSubscription`.
- `DELETE /profiles/{id}/push-subscriptions` returns 204.
- Both are in the row-scope matrix (`NOT_WALKED`).
- `GET /deployment` now also says `push_key`, the key the app subscribes with, or null.

**Which sender a process runs on (#123):**
- The VAPID keys set: `WebPush`, for a real deployment, a demo, or a laptop that wants it.
- No keys: `FixturePush` on a dev run; `NoDevices` in a demo, which reaches nobody.
- A real deployment without keys still refuses to start, because `NoDevices` is a fixture.

**Settings (`docs/deploy.md`):**
- `NURA_VAPID_PUBLIC_KEY`, `NURA_VAPID_PRIVATE_KEY` (secret) and `NURA_VAPID_SUBJECT` come from the environment, never the repo.
- The three go together or not at all, and a pair that doesn't match refuses to start.
- The provider table and a one-line key generator are in `docs/deploy.md`.

**CI image job:** a new step makes throwaway keys inside the job. With them, the demo's push is `WebPush` (not a fixture), and production's check no longer names `push` among the fixtures it refuses.

**Data map:** the new table's columns are classified, and `docs/trust/pdpa-data-map.md` is regenerated.

## Dependencies (exact pins)

- `cryptography==50.0.1`
- `http-ece==1.2.1` (requires `cryptography>=2.5`)
- `py-vapid==1.9.4` (requires `cryptography>=46`)

The only other packages they pull in are `cryptography`'s own: `cffi` 2.1.1 and `pycparser` 3.0. Nothing unexpected. No existing pin is changed.

## Web

- **`web/src/sw/push.ts`:** the push handler in its own service-worker module. It shows the one line and nothing else, and a tap opens `/app/?card=<id>`. `sw.ts` gets one import line, so W4's offline work there doesn't meet it.
- **`web/src/push/reminders.ts`:** the browser's push APIs are passed in, so tests can stand in for them. `requestPermission` is called only in `turnOn`, which only the tap calls, never on load.
- **Me:** "Get reminders on this phone" and "Stop reminders on this phone", shown only when the deployment has a push key and the browser can push. If the phone said no, it says so in two lines. The strings are in en/ms/zh and pass `make plain-words` and `make language`. The tile sits below switch profile / set up, before the home-screen hint and sign out, well clear of #138's proud number, which is the first tile on Me.

## Tests

`backend/tests/test_web_push.py`:
- **RFC 8291 Appendix A:** `encrypt_payload`, through `http_ece`'s deterministic path (the example's salt and sender key), gives the published body byte for byte: the 86-octet header and the published ciphertext. The example's receiver keys decrypt it back. The body is 144 octets; the RFC's example `Content-Length: 145` line is off by one, and the body it prints is 144.
- **VAPID:** the token's `aud` is the push service's origin, its `exp` is at most 24 hours ahead, and it verifies with the public key it names. A key pair that doesn't match is refused at start.
- **Settings:** the three VAPID settings go together, and `mailto:` or `https:` is required. The sender choice is checked for each case above.
- **Behind the trigger engine**, against a push service stood in with `httpx.MockTransport`:
  - every push to Mei's phone is one aes128gcm request that decrypts to exactly a line and an id, with no health word (not the medicine, not the count, not what it is about), and the reorder's push names its delivery row;
  - the quiet hours hold it, with no request made;
  - a 410 drops the subscription, WhatsApp takes that send, and it isn't tried again;
  - a phone signed out gets nothing;
  - a nudge's push carries the nudge's own id;
  - a nudge he answered in the app gets no push.
- **Over HTTP:** subscribe, subscribe again (which replaces), a refused http endpoint, then unsubscribe.

`web/tests/unit/reminders.test.ts`: reading the state never asks the phone; a tap asks once and subscribes with the key bytes; a no subscribes nothing; stopping forgets the subscription on both the backend and the browser.

`web/tests/e2e/push.spec.ts`: Playwright generates throwaway VAPID keys at config load, so the e2e backend runs Web Push and no key is committed. The push service is stood in. The test checks:
- Me shows the button and the phone hasn't been asked;
- a tap asks once, the backend answers 201, and it shows "Reminders are on for this phone.";
- Stop answers 204 and the button is back.

## Checks

On main at `3d4d511` (#138), merged into this branch by that exact SHA. The only conflict was `web/src/screens/Me.tsx`'s imports.

- **Backend:** `python3 -m pytest -q -p no:cacheprovider` gives 1427 passed, 1 skipped (SQLite). `backend-postgres` in CI runs the same suite on Postgres.
- **`make lint`, `make plain-words`, `make language`:** green, 0 failures.
- **Web:**
  - `npm test`: 23 files, 195 tests pass;
  - `npm run build`: passes, and `sw.js` carries the push handler;
  - `npm run plain-words`: 0 failures;
  - Playwright: `push.spec.ts`, `today.spec.ts`, `offline.spec.ts` and #138's `day.spec.ts`, 22 of 22 pass, against a backend on port 8119 started by Playwright with throwaway VAPID keys.
- **Checkpoint 20** (delivery): 16 steps pass, on a fresh DB with a frozen clock on port 8119. The server was stopped by its PIDs.
- **`alembic check`:** clean on the new migration (`tests/test_models_registry.py`). `docs/trust/pdpa-data-map.md` is regenerated.


ashfrnndz21:
Operator merge note: CI backend, backend-postgres, image and web green on 705102e (main through 3d4d511); the GitHub reviewer is down, so the operator reviewed in its place. Operator decisions: vetted libraries for the payload (http-ece, py-vapid on cryptography, exact pins, no requests), not hand-written crypto; the push port takes the run's session, key context and a payload id (one caller, delivery). Accepted: RFC 8291 Appendix A reproduced exactly; VAPID JWT audience and 24h expiry verified; a payload of one fixed line and an id, never health words; #121's caps and quiet hours; 404/410 drops the subscription and falls back to WhatsApp; a signed-out phone gets nothing; an answered nudge is never pushed; permission asked only on the tap; the push handler as its own service-worker module; a real deployment still refuses to start without VAPID keys. Builder ran the suite (1427, 1 Postgres-only skip), lint, plain-words, language, web unit (195), e2e (22), checkpoint 20. Follow-up: a push opening its card, not Today.

PR #165, opened 2026-09-15
E21-01 fix: the feed's fresh first page merges in below the card on screen, never lost

ashfrnndz21:
## What this fixes

On main, `day.spec.ts:521` ("the post-visit card on the web") fails in CI. It is a race in the feed pager (E21-01).

The pager opens on the page the phone kept, then asks for a fresh first page. If the fresh page landed after he had scrolled past the first card, it was kept for next time and not shown. A card made in between was lost from the screen, and the pager could run out under his finger. The post-visit memo card is one such card.

## The change

**`web/src/feed/store.ts`.** A fresh first page that lands after he has read on is now merged in below the card on screen:
- Every card up to and including the one he is on stays exactly where it is.
- The fresh page's cards he has not passed follow, in the backend's order. A card made since (a flag, the memo card) is the next one down.
- Paging continues from the fresh page's cursor.
- A page of the old list that lands after the merge is dropped, by generation.
- Card keys: a whole list (the kept page, or the fresh page on the first card) keys its cards `0:n`, as before, so the fresh page taking the first card's place keeps the same elements — a focused card keeps its focus. A merge, and every page after it, counts on from 1, so no key repeats. (The first push gave every list new keys; the remount dropped focus and Page Down did nothing — CI's `feed.spec.ts` "nothing covers a line" at 360 by 640. Fixed in `dd8315d`, with a unit test.)

**`web/tests/unit/feedStore.test.ts`.** Unit tests for the merge. What he has passed and the card on screen don't move, the new card is next, paging goes on from the fresh cursor, and a stale page is dropped.

**`web/tests/e2e/day.spec.ts`.** The post-visit walk waits for Today's feed response, then for the pager's own fresh first page, before paging to the memo card. It no longer races it.

Nothing else changes. No backend change, no new strings, and the client composes no sentence.

## Tests

Run in the worktree, on a merge of main at `6467623` (exact SHA), against this worktree's own server on 8132.

On `dd8315d`:
```
npx vitest run                                              245 passed (245)
npm run build                                               ok (tsc strict + vite)
npx playwright test tests/e2e/day.spec.ts --repeat-each=5   70 passed (4.2m)   (fresh dev.db)
npx playwright test tests/e2e/feed.spec.ts visit.spec.ts    12 passed (1.3m)
npx playwright test                                         76 passed, 1 failed (17.2m)
```

`day.spec.ts` passed 5 of 5 runs on a fresh database, every test in the file each time, including `:521`.

Three failures came from the test setup, not this change:
- **The full run's one failure** was `visit.spec.ts:32`: a recording timer read 1:16 where the test expected 1:06. That run took 17.2 min instead of 5.9, because the Mac was busy with other suites. On a re-run, `visit.spec.ts` passed.
- **An earlier ×5 run reused `dev.db` after four runs.** It hit `409 ProfileAlreadyOwned` in the seed: the e2e phone numbers are unique only within one run. It also hit the known stale read in `Symptoms.tsx` (`:409`). On a fresh database, all 70 passed.

On `f9e8cc8`, the first push: 244 unit tests, the full e2e 77/77, and `day.spec.ts` ×5 70/70.


ashfrnndz21:
Operator merge note: web-only (feed store, its unit test, day.spec's wait) and risk:low; backend, image and web green on dd8315d, and the web job includes the day.spec walk this fixes. Merged ahead of backend-postgres (a backend suite this change cannot affect) so every other builder's e2e stops tripping on the race. Accepted: the fresh first page merges into the stream below the card on screen, never moving what he is looking at.

PR #169, opened 2026-09-15
Red flags reach someone whatever the settings; no medicine instructions in family messages; WhatsApp consent narrowed

ashfrnndz21:
Closes #163, closes #164
Refs #162

Fixes from the retro review of #121, #137 and #151. Other builders own the WhatsApp "Taken" part of #162 (R2) and every string rewrite listed in the three issues' tables (R1), so this PR leaves them alone.

## What changed

**1. No setting can silence a red flag** (#162 check 5, #164 check 5)
- Red-flag delivery is not configurable. `check_settings` refuses a `flag` channel list (`AlertsGoEveryWay`, 400), the same way it already refuses a cap.
- A `flag` list or cap stored before this change is ignored (`config_of`, `Config.channels_for`). There is no migration.
- Every rung tries every way each person can be reached: an app push if they have a device, and WhatsApp where they may be sent it. There is one `Delivery` row for each channel that carried it.
- The notice on their family page is always written: a new `DeliveryChannel.IN_APP` row. That row is what shows them the open flag and its "I'm on it" button.
- When no phone was reached, a `NO_CHANNEL` row says so, and the ladder moves on at once (`ladder.climb`). The next rung is due immediately, and later rungs keep their gaps from there. This is recomputed from the rows on every run.
- The chief sees who could not be reached:
  - `GET /ladders` → `not_reached`: "Nura could not reach Siti on their phone." (owner and chief only)
  - `GET /profiles/{id}/reach` (owner and chief; registered in the row-scope matrix): "Nura cannot send Siti messages on WhatsApp." and "Siti sees what Nura says only in the app."
- `Escalated.told` now counts only phones reached. The inbound "Mei knows now." and the not-feeling-well and symptom-log cards name whose phone was reached, else who was asked.
- The web Delivery screen shows the red-flag row read-only: "Nura always tells your family about this, every way it can." (en, ms, zh). Saving never sends a flag row.

**2. Messages to Pa can't carry medicine instructions** (#164 checks 3/4)
- `water_pill_morning` is gone from `family/strings.py`, `web/src/family/model.ts` and the web labels.
- A message to him that names a medicine or a dose is refused (`MessageNamesAMedicine`, 400), as E03-03 refuses a provider note.
  - The check is `health_words.names_medicine_or_dose`. It uses the licensed registry through its port, the high-risk table, his names for medicines, the plain medicine words in all three languages, and an amount with a unit, in digits or words.
  - It applies to templates, slots and memos, at preview and again at schedule.
[...]

PR #168, opened 2026-09-15
WhatsApp inbound: an unheard voice note reaches a person; the webhook retries failures, once per message; Taken asks which tablet

ashfrnndz21:
Closes #158
Refs #162

## What changes

**#158: a voice note Nura could not hear is no longer a dead end**
- If the transcriber heard nothing, the transcriber was down, or the audio could not be fetched, Pa's reply ends with "If you feel unwell, call your family now." in all three languages.
- His chief is told on her own channels through a new trigger, `VOICE_NOTE_UNHEARD`. If her key opens his notes, the `unheard_note_notice` template tells her to listen in the app or call him. If it doesn't, or the note could not be fetched, `unheard_note_notice_call` tells her to call him. The notice carries none of the note's words or audio, and the note stays private.
- **Why ALERT.** A note with no words heard in it may hold a red word that nobody could read. It is treated the way a red flag's notice is:
  - `Category.ALERT` under `Scope.EMERGENCY`, with `cap=None` and `quiet=False`, so no cap or quiet hours can hold it (`check_settings` refuses a cap).
  - It goes on the rule's own channels (WhatsApp, then the app's content-free push), passed explicitly, so a stored channel setting cannot redirect it.
  - Every attempt is a `Delivery` row.
  - A false alarm costs the family a call. A missed red word could cost a fall that nobody hears about.
- Both templates are `approved=False` until Meta approves them. A dev run sends them. Until then a deployment falls through to the app push.

**#158: the webhook no longer swallows failures**
- The new `whatsapp_receipt` table records each message by the provider's id, with a unique constraint and never the words. Migration `0030_whatsapp_receipts` sits on main's head, `0028`; B1 takes `0029`. The table is classified in `scripts/data_map.py` and the data map is regenerated.
- Each message is handled in its own savepoint (`receipts.receive`). Any failure answers 503, so the provider sends the delivery again. Messages already handled are skipped by id. Failures are counted by id and by exception class name, and from the 3rd failure each one is logged by the receipt's id.
- Red flags are still read first, before "ignore" and before the consent check (`handle_inbound`, unchanged).
- Once a red flag's ladder has run, a reply that fails afterwards is logged. It no longer rolls the flag back, which would have told the family again on the retry and never kept the flag.

**#158, non-blocking items**
- `group_for` is replaced by `app/keys/handles.profile_for_group`. It writes a system READ line on the family's trail.
[...]

Issue #163, opened 2026-09-15
Retro review of #151


ashfrnndz21:
Retro review of #151 (consent: family alerts under their own keys, close my account, a push opens its card), merged while the automated reviewer was down. Both checklists (`.claude/agents/clinical-safety-reviewer.md`, `.claude/agents/plain-words-reviewer.md`) were applied by hand, and every finding was checked against `main` at 6467623. `make plain-words` and `make language` report 0 failures on main.

## Clinical safety: 7 pass, 1 fail

| # | Check | Result | Where on main |
|---|---|---|---|
| 1 | Profile reads through the keys module | PASS (note) | a closing profile is refused at the door, on the trail (`backend/app/keys/context.py`, `AccountClosing`, `only_the_owner_while_closing`). Note: `answerable_while_closing` reads `Ladder` and `Flag` raw (`backend/app/identity/closing.py:246`, `:250`). Nothing reaches the caller, but the row-scope test only catches Fact, Artifact and Event, so nothing guards these reads |
| 2 | Facts carry provenance; cards record State | PASS | the consent rows are archived to `ErasureRecord` before the graph goes (`identity/closing.py:304-310`, `:370-379`) |
| 3 | No output changes a medicine | PASS | not touched |
| 4 | Drug data from the licensed client only | PASS | not touched |
| 5 | Red flags escalate before ranking or caps | PASS | family red-flag notices no longer wait on his WhatsApp agreement (`backend/app/channels/whatsapp/outbound/send.py:137-161`). A closing account still escalates the flags raised before it (`backend/app/delivery/triggers/engine.py:111-116`), and the closing words tell him his family will not be told, and who to call (`backend/app/consent/closing_words.py:20-21`) |
| 6 | High-risk drugs need a label photo | PASS | not touched |
| 7 | Health data in region; none to analytics | PASS | deletes stay under the region's prefix (`backend/app/ingestion/s3.py`, `delete_prefix` / `check_prefix`); a push carries only an id (`web/src/sw/push.ts`, `web/src/push/open.ts`) |
| 8 | Consent at claim and per key; refusals on the trail | **FAIL** | see below |

**8 FAIL: stopping WhatsApp stops less than his confirm lines say, and a key holder's "no" to WhatsApp is ignored.**

[...]

Issue #158, opened 2026-09-15
Retro review of #146


ashfrnndz21:
Retro review of #146 (B2: transcript search, notes in recall, story and learning cards, WhatsApp family group and Pa's voice notes, webhook cap), merged while the automated reviewer was down. Both checklists (`.claude/agents/clinical-safety-reviewer.md`, `.claude/agents/plain-words-reviewer.md`) were applied by hand, and every finding was checked against `main` at 6467623. `make plain-words` and `make language` report 0 failures on main; the string failures below are ones the lexical verifier cannot see.

## Clinical safety: 6 pass, 2 fail

| # | Check | Result | Where on main |
|---|---|---|---|
| 1 | Profile reads through the keys module | **FAIL** (non-blocking) | `backend/app/channels/whatsapp/group.py:83-89`, see below |
| 2 | Facts carry provenance; cards record State | PASS | `backend/app/ingestion/notes.py:313` (his voice note is an `OWN_NOTE` artefact on its own event); `backend/app/delivery/feed/compose.py:1305` (story cards name their facts) |
| 3 | No output changes a medicine | PASS | `compose.py:1410` (a story leaves medicine memos to the latest memo card) |
| 4 | Drug data from the licensed client only | PASS | `backend/app/channels/api/medicines.py:262-266` (the story's voice comes from the registry's story) |
| 5 | Red flags escalate before ranking or caps | **FAIL** (blocking) | see (a) and (b) below |
| 6 | High-risk drugs need a label photo | PASS | not touched |
| 7 | Health data in region; none to analytics | PASS | `backend/app/channels/whatsapp/inbound.py:1118` (voice heard by the region's transcriber); `backend/app/search/transcripts.py:144`; logs carry class names only |
| 8 | Consent at claim and per key; refusals on the trail | PASS | `group.py:169` (WhatsApp agreement before the group is filled or mirrored); `group.py:233` (only owner, steward or chief opens it); `backend/app/memory/episodic.py:183` (family photo rests on HOLD_HEALTH_RECORD); `transcripts.py:142` (family-only, refused on the trail) |

[...]

PR #180, opened 2026-09-16
WhatsApp: any key-holder's unheard note reaches a person; a failed channel falls through; receipts claimed first

### LINKED ISSUE #173: WhatsApp follow-ups from #168: unheard notes from the helper or without consent; failed WhatsApp sends fall through to push
Follow-ups the builder of #168 listed; the first two are safety-relevant:
1. **An unheard voice note from the helper (Siti), or from a patient without WhatsApp consent, tells nobody.** Treat every unheard inbound note the same way: the chief is told on her own channels (alert category, as #168 decided) and the note is kept for a person to hear.
2. **A WhatsApp send that fails with a network error stops instead of falling through** to app push and the in-app notice. For every trigger, a failed channel falls through to the next reachable one (and for alerts, #169's every-way rule already writes the in-app notice — verify it holds on a send error).
3. The unheard-note notice has no escalation ladder: if the chief does not acknowledge, the next rung should be asked, like a red flag's ladder.
4. His reply could say 'Mei knows now.' once the notice actually went out.
5. Two copies of one delivery at the same instant get a 500 and are re-sent (nothing lost) — make the second one a clean duplicate acknowledgement.
6. English 'can see you took it' reads oddly after two tablets.
Also: the two notice templates need Meta approval; until then, outside the 24-hour window the chief gets only app push (and the in-app notice).

ashfrnndz21:
Closes #173.

The three safety follow-ups from #146, #158 and #168, in the WhatsApp inbound and delivery paths.

## 1. Any key-holder's unheard voice note reaches a person

`_voice_note` / `_voice_not_heard` were gated on the sender being the owner, so a note from Siti went to `_other()` and told nobody, and a note on a profile whose patient has not agreed to WhatsApp was answered with the consent refusal and told nobody either.

Every voice note is now heard in region whoever sent it — that was already true, and the red-flag check still runs on its words before "ignore" and before the consent — and one Nura could not hear always reaches a person:

- **His own**: unchanged, except that the reply now names who the notice reached.
- **Anyone else's**: not kept (it may carry other people's voices, which only a consult keeps), the refusal on the trail by name, the chief told on her own channels as an alert. The notice is a new template, `unheard_note_notice_from`, which names **who actually sent it** and says to call them — a clinical-safety review caught that the existing notice would have told Mei "Pa sent a voice note" when Siti had. `delivery_ladder.note_from_person_id` (migration 0031) is how the ladder remembers that.
- **His line stays his**: "If you feel unwell, call your family now." is said only where the sender is the patient. Siti is asked to write what she said, and told who to call if it cannot wait.
- **No WhatsApp agreement**: nothing of the note is kept — no artefact, no thread, no message row — the sender gets one fixed line straight from the provider, and his family is still told, **through the app**. A red flag remains the one thing his family is sent on WhatsApp after he stops it (#163): `rules.told_without_his_agreement` now says so in one place, and `deliver._no_whatsapp` uses it instead of `is_alert`, so the send door and the delivery door finally agree. No consent wording was touched.
- **The family's group** is out of it: a voice note posted there is the family's, the way a photo there is — not kept, never an alert. A red word in one is still a flag.

## 2. A failed channel falls through to the next reachable one

`_by_whatsapp` caught only `Refusal`, so a network error stopped the whole delivery. Every channel attempt now runs through `_attempt`: a door's no keeps its trail line and moves on, anything else rolls that channel's savepoint back (dropping the keepers registered inside it), writes the failure on the row's `passed_over` by the name of its class, and tries the next channel. Same for the app push and for the caregiver stand-in.

#169's every-way rule is proved on a send error: a red flag whose every channel failed writes a `NO_CHANNEL` row **and** the in-app notice, and the flag and its ladder stand. There is also a test for the case the issue names — the two notices are pending Meta's approval, so on a real number the send door refuses them and the notice goes by app push and the family page.

## 3. The receipt row is claimed before the handler runs

`receipts.receive` now inserts the receipt first, under the unique constraint on the provider's id, in a savepoint of its own. Two copies of one delivery at the same instant: exactly one runs the handler, and the other loses the race and is answered as the duplicate it is instead of with a 500. A row an earlier failure left behind is claimed with `SELECT … FOR UPDATE`, so a redelivery racing another redelivery waits and then finds it handled.

## 4. The telling is contained, and a real failure is retried

The issue asked for a catch-all around `_tell_family_unheard` so a failure there never rolls back the reply already sent to him. **I implemented the intent rather than the letter, and want that flagged.** The notice now goes *before* the reply (item 5 needs it to, so the reply can name who knows), so at the moment a failure can propagate nothing has been said to the sender. That makes the catch-all the wrong tool: it would answer the webhook 200 with nobody told about a note that may carry a red word — the one outcome #173 exists to prevent. So:

- `_tell_unheard` catches `Refusal` (a door's decision, on the trail, not retried) and lets anything else go on up: `receive` records the failure, the webhook answers 5xx, the provider sends the message again, and the retry tells him once — not twice. This is what `_escalate` does for a red flag.
- `_say_unheard`, the reply, keeps the catch-all: a lost reply is not a safety loss and must never roll the note or the notice back. This is what `_red_flag` does for its reply.

Both are covered by tests.

## 5. The smaller items

- The notice climbs a ladder of its own (`Subject.UNHEARD_NOTE`, `UNHEARD_RUNGS`): the chief at once, whoever is on duty five minutes later, then everyone else whose key holds the emergency card. One person's "I'm on it" stops it for the rest, through the same door a red flag's ladder is answered through, and the engine's five-minute run is the net under it.
- His reply says "Mei knows now." once the notice actually went — and "Mei and Kit know now." for more than one, and nothing at all where it reached nobody. Nobody is ever said to know who was not told.
- A duplicate delivery is a clean acknowledgement (item 3).
- "Mei can see you took them." after two tablets.

## Gates

`make test` (SQLite: 1641 passed, 1 skipped), `make lint` (ruff + mypy over 252 files), `make plain-words` (0 failures), `make language` (0 failures), `python3 -m scripts.data_map --check`, and checkpoint 9 against a live dev run on port 8137. New migration `0031_unheard_note_ladder` sits on main's head; `alembic heads` is one head.

Note: `tests/test_whatsapp_inbound_safety.py` was **red on main** before this branch (three failures from #168: `[notice] = await _notices(sg)` unpacking two rows since #169 made an alert write the in-app notice too, and a `change()` call that an alert now refuses). Those are fixed here.

## Known follow-up, not fixed here

The loser of the receipt race answers `ALREADY` while the winner may still fail and roll back. Nothing is lost today — the winner's own HTTP attempt gets a 5xx and is redelivered — but answering `FAILED` on the lost race would make the "never a 200 for a message we did not handle" rule hold without depending on that.

---

## Clinical-safety review

`.claude/agents/clinical-safety-reviewer.md`, run twice — once on the first draft, once after its findings were addressed.

**First pass** found three FAILs and six required changes. In summary:

- **Check 2 (provenance) FAIL** — the unheard-note notice filled `{name}` with the patient's name on a path now reached for any key holder, so Mei would have been told *"Pa sent a voice note to Nura… Call Pa now."* when Siti had sent it. "The family is given a false account of who did what, and acts on it by calling the patient."
- **Check 5 (red flags escalate first) — PASS on ordering, FAIL on the failure mode.** Detection order verified correct for every sender: `_hear` runs for any voice note, `detect` runs before the ignore check, before `_whatsapp_agreed` and before `require_consent`, and the new unagreed branch is guarded by `not flagged`. But the catch-all around the telling meant an infrastructure failure would answer the webhook 200 with nobody told — "less safe than the flag path it is modelled on".
- **Check 8 (consent) FAIL** — `TOLD_ANYWAY` widened what the family is sent on WhatsApp without the patient's standing agreement, which is not what `STILL_TOLD` / `NOTHING_ELSE` promise him when he stops. "The widening is not required to satisfy #173's 'a person is always told'."
- Plus: the family's WhatsApp group was not excluded, an alert-fatigue source; `_attempt` should trim keepers from a rolled-back savepoint; and the receipts race.
- Checks 1, 3, 4, 6, 7 PASS. On locality: "the helper's audio is never fetched into storage… `note_id` is always `None` on that path… The notice carries no word of the note."

**Second pass, after the fixes:**

> Gates re-run green: `pytest tests` 1639 passed / 1 skipped, `ruff check .` clean, `mypy app` clean over 252 files, `app.language.memory` 0 failures, `app.safety.plain_words` 0 failures.
>
> **1. Attribution — CLOSED, with a residual.** `unheard_note_notice_from`, `delivery_ladder.note_from_person_id` and the `_from` selection fix the real path; the test now pins `template_name == "unheard_note_notice_from"`. Residual: `ladder.py:526` falls back to the **patient's** name when `run.person()` returns None, and treats `run.patient is None` as from-patient — both re-create "Pa sent a voice note to Nura about Pa… Call Pa now."
>
> **2. Failure retries — CLOSED.** `except Refusal` at `inbound.py:1651`. I verified the ordering argument independently rather than taking it: `_voice_note`, `_voice_not_heard` and `_note_unheard_from_someone_else` all run the telling before `_say_unheard`, and `_keep_voice` rolls back with the receipt savepoint, so a retry keeps the note once and tells him once.
>
> **3. #163 — CLOSED.** `told_without_his_agreement` and `deliver._no_whatsapp` now agree with the unchanged `RED_FLAG_NOTICES`; no consent words were touched, so nothing promises more than it did before.
>
> **4. Group notes — PARTIALLY closed.** [the `_unheard_unagreed` branch had no group check]
>
> **5. Keepers — CLOSED.** Scoped to the rolled-back range rather than clearing the list.
>
> **6. Receipts race — accepted** as a documented PR follow-up; nothing is lost today.
>
> Nothing new beyond the above. The sender's display name now crosses to every emergency-card holder, which is required for the notice to be true and stays inside the same family — no action.

Both remaining points are fixed in `15e0b2b`: the notice now raises `SenderNotNamed` rather than guess (the channel falls through to the push, and the family-page notice is written all the same), and `_unheard_unagreed` is gated on `group is None`. Each has a test.

---

## Plain-words review

`.claude/agents/plain-words-reviewer.md`, run twice.

**First pass** — one FAIL, eight PASS:

> **`voice_note_unheard_told` — PASS.** `{names} knows now.` is exactly the glossary's wording for *escalated/notified*, and `sudah tahu` / `已经知道了` are the §6 rows.
> **`voice_note_unheard_told_many` — PASS.** `{names} know now.` is the right plural and matches `red_flag` against `red_flag_one`. Malay and Chinese are identical to the singular, which is correct — neither marks number on the verb.
> **`voice_note_not_fetched_told` / `_told_many` — PASS.**
> **`note_unheard_other_told` — PASS.** The safety line that belongs to the patient is correctly absent.
> **`note_unheard_other_told_many` — PASS.**
> **`note_unheard_fixed` — PASS.** Follows `closing` / `group_not_kept` / `red_flag_fixed` exactly.
> **`taken_patient_many` — PASS.** `took them` is right, and the `{took}` line above it names the tablets, so "them" has an antecedent.
>
> **`note_unheard_other` — FAIL.** This is the one sent when the notice reached **nobody**. Siti is told Nura could not hear her and asked to write — and that is all. She is not told whether anyone knows, and she is given nothing to do if it cannot wait. Every sibling in the catalogue for the same situation gives a route… That is rule 7 (say who does the next thing) and rule 9 (the small reassurance that removes a worry).

The rewrite was applied verbatim ("If it cannot wait, call the family now." / "Kalau tidak boleh tunggu, telefon keluarga sekarang." / "如果不能等，现在就打电话给家人。"), and, as advised, the two `_told` variants were left alone.

Two notes, neither a failure: the knows-now line sits before the action line here and last in `red_flag` — a deliberate divergence, since ending on the action is the stronger last line; and `mendengar` vs `dengar` in the Malay catalogue is pre-existing on main, faithfully reproduced rather than half-fixed.

**Second pass**, on the new template added for the clinical finding:

> **1. `note_unheard_other` — PASS.** My rewrite is in verbatim in all three languages; the two `_told` variants are untouched.
> **2. `UNHEARD_NOTE_NOTICE_FROM` — PASS.** It matches its neighbours word for word where it says the same thing: line 2 is byte-identical to both `UNHEARD_NOTE_NOTICE` and `UNHEARD_NOTE_NOTICE_CALL` in en/ms/zh, and line 3 is the `_CALL` variant's closing line with `{who}` swapped for `{name}`. Line 1 extends the shared opener with "about {name}" — 9 words filled, natural in all three, correct Chinese word order. It rightly copies the `_CALL` shape rather than the base one (nothing of her note is kept, so there is nothing to listen to in the app).
>
> One follow-up, not a string issue: the docstring of `_note_unheard_from_someone_else` still says the chief "is asked to call him"… the docstring is now stale and should read "call whoever sent it".

The docstring is fixed in `ccf7f9b`.


ashfrnndz21:
Operator merge note: CI backend, backend-postgres, image and web green on 248bc49 (main e76c223 merged, migration renumbered 0032_unheard_note_ladder on 0031_brief_safety, one head). The repo's reviewer is out of API credits, so the builder's two self-reviews and an independent review (verdict MERGE, no blocking, each claim traced to code and tests) stand in. On the merge it took main's is_red_flag_notice wholesale — the unheard-note notices deliberately do not share that prefix, so they never go out without his WhatsApp agreement — and found one real defect its own merge introduced: #147's _notices helper now excludes the in-app row, so five assertions were reading the wrong set; fixed to read every row. Closes #173.

PR #179, opened 2026-09-16
Add someone and grant access, and they can ask at once — proven end to end

ashfrnndz21:
## Summary

The owner's own rule, plainly: on the Family screen, the main user (the patient, the profile's owner) can add and grant access to a person, and the moment he does, that person can query his profile, its context and its memory — over their own key, on their own account. This PR proves that path end to end and fixes what it found broken.

## What was broken

1. **The owner had no way to reach the screen at all.** The Family "Give someone a key" screen (`web/src/screens/family/Keys.tsx`) was only ever listed in the caregiver density's Family home (a chief's) — never in the patient density's, which is what the owner himself sees by default. `GET /profiles/{id}/keys`-style access was there in the backend the whole time; there was simply no selection to reach it as the owner. Fixed in `web/src/screens/family/Family.tsx`: the owner's own Family home now offers "keys" too.
2. **Where it was reachable, adding a genuinely new person always failed.** `NewKey` in `Keys.tsx` called `POST /profiles/{id}/keys` straight away. That route refuses with `ConsentWithheld` unless a `SHARE_WITH_PERSON` sharing agreement is already in force for that person — and letting someone in at all is the *owner's own yes* (`may_invite`), never a chief's, so a chief cutting a key for someone the owner never introduced hits the same wall. The onboarding invite gap (`Invite.tsx`) already did this correctly (preview the words → agree → cut the key); `Keys.tsx` never got the same treatment. Fixed: when the owner is using the screen, he now reads the backend's own rendered words for exactly the parts and the role (`POST /consents/sharing/preview`) and agrees to them (`POST /consents/sharing`) before the key is cut. A chief's path is unchanged — she can still only cut a key for someone the owner already let in.

Everything else in the path — resolving a key context fresh on every request (no caching, no job), `GET /doors` reflecting a grant or a revocation at once, `POST /ask` requiring the `ask` scope (which the caregiver role already holds by default — the one "likely suspect" from the brief that turned out *not* to be a bug), recall answering with citations in the caller's own explicitly-passed language, and refusals being written to the owner's trail — was already correct and is now covered by tests that would have caught a regression.

## Proof

[...]

PR #170, opened 2026-09-15
Retro review fixes: a medicine is never blamed without 'do not stop it', Sad not Low, refusals on the trail, strings

ashfrnndz21:
Closes #156, closes #157, closes #159, closes #160, closes #161

Refs #162, refs #164

The fixes from the retro reviews of #115, #119, #121, #123, #125, #137 and #138. It started from the operator's save (`29de47c`, backed up at `wip/fix-retro-reviews`) and was finished from there.

## What changed

- **#156, invite refusals on his trail.** `HolderNeedsAName` and `NoSuchHolder` in `let_someone_in` and `preview_letting_in` (`channels/api/profiles.py`) now run inside `audited_guard`, so each one is written on the owner's trail. The preview checks the family scope before it looks anyone up. Tests: both refusals on both calls (`test_sharing_preview.py`). #156's strings: `onboarding.cloud.noted`, `onboarding.plan.missing` (now a whole line, tagged `@patient`) and `onboarding.plan.nothing`.
- **#157, a medicine is never named without "do not stop it".** When the feeling note says "This can come from {medicine}, new since {date}.", the next two lines are "Do not stop {medicine} yourself." and "Tell {doctor} how you feel.", in en, ms and zh (`DO_NOT_STOP`, `inference.py`).
  - Rule 14 of the plain-words verifier refused any "stop" next to a medicine. It now has one narrow allowance, `_KEEP_TAKING`. That allowance passes only this exact shape: one treatment verb, no digit, no punctuation but the full stop, and no word that adds a condition or a second thing. The clinical reviewer's probes are now failing cases in `test_plain_words.py`.
  - The wording is row 2 of `docs/trust/clinical-wording-sign-off.md`, next to #126's body salt. The row names the watch-outs these lines can follow.
  - "Low" is now "Sad" in English: the cloud word, the TELL line and the `FEELINGS` phrase. Malay and Chinese already said sad.
  - **The hypoglycaemia red flag stays reachable.** "Low" used to be the only nearby word on every cloud, so a man on insulin or a sulfonylurea could tap it for a low sugar. Now "Shaky and sweaty" stays on his cloud for as long as he takes such a medicine, not only in its first fortnight (`SUGAR_MEDICINE`, the red-flag module's own `HYPOGLYCAEMIC_CLASSES`). A tap on it takes the red-flag path, unsuppressed, on day 15 too (`test_feelings_cloud.py`).
[...]

PR #166, opened 2026-09-15
Record review fixes: ask-to-order confirmed once a day, the family task names Pa, high-risk counts need a photo, 52 plain-words rewrites

ashfrnndz21:
Fixes the three **blocking** findings from the #140 review, plus a race the follow-up review of this branch asked to close before merge.

## The #140 review's blocking items, and the fix for each

**1. "Ask the family to order." wrote a task and queued a notice with no confirm — and nothing stopped two taps making two tasks.**
`reorder.py` now mints an `OrderDraft` (`ConfirmSubject.ORDER`) naming the line and the person the preview showed, and `ask_to_order` consumes it before anything is written — the same shape as `TaskDoneDraft`/`DriveDraft`. `order_preview` reads who is on duty now (whoever holds a key that opens the medicines, else the chief) and, when the family was already asked for this line today, says so and names who instead of a fresh ask. A second yes that day answers with the task already on the list — no new task, no new notice.

[...]

PR #147, opened 2026-09-14
B1 The brief at T-3, symptoms in it, a two-language emergency card, the call-clinic row, routine-anchored doses

ashfrnndz21:
The T1 parity audit found backend gaps in the pre-visit brief and the safety tables. This PR closes them, adds two things the coordinator asked for on the way, and acts on both self-reviews. It is clinical work, so `risk:high`, and the red-flag tiers are marked as awaiting a clinician's sign-off (ADR 0010).

### E05-01 The pre-visit brief (#33)
Acceptance: *Generated at T-3 days; editable; one page.*

- **Generated at T-3.** A new rule trigger in the E11 engine renders the brief three days before a visit and hands its card to delivery.
  - Type `visit_brief`, rule `brief_three_days_before`, one a day, quiet hours held.
  - It fires from his breakfast on. A visit booked inside the three days gets its brief the first run it can, until the day before (that day is the visit reminder's).
  - The brief row is written then (`built_at` is T-3), and the delivery row names it.
  - The WhatsApp card is a new pending template, `visit_brief`. Its slots are fixed (`doctor`, `day`, `time`, `subject`) because a Meta template parameter cannot hold a line break.
  - A day the brief reached him counts as the visit's one reminder, so the anticipation nudge is skipped that day.
  - The visits guard now lets the engine's system context render the brief and refresh its questions, and nothing else (`guard.may_render_brief`).
  - Test: `tests/test_brief_b1.py::test_the_brief_is_rendered_three_days_before_and_its_card_goes_once_under_the_cap` (T-3 once, the second visit held by the cap and sent at T-2, nothing on T-1).
  - Checkpoint 20 has a new step for it.
- **One page.** `LINE_BUDGET = 21`: purpose 2, what changed 4, symptoms 4, questions 4, bring 4, boundary 3.
  - A section that would run over keeps what fits and ends on one line saying how many more ("Nura has 6 more questions for Dr Tan."). The count is always two or more.
  - A first entry too long for the page is cut to fit, never dropped.
  - Test: `test_the_brief_fits_one_page_whatever_the_record_holds`.
- **Editable.** The operator decided the brief's lines come from the record and are **not** edited. What he and his chief want to add goes on the questions card (E05-02), which is editable on a yes. This is written in `brief.py`'s docstring and in a comment on #33.
- Also: counts of one are now singular ("1 new number is in your blood pressure book.").

### E14-01 Symptoms in the brief (#57)
Acceptance: *Logged in his words; appears in the pre-visit brief.*

[...]

PR #190, opened 2026-09-16
A chief reads his notes again

ashfrnndz21:
## What this is

Checkpoint 13's step 4 was reported failing on a clean checkout — "Mei reads Pa's notes as his chief" — after today's run of #169, #147, #180, #170, #179. That would be a real regression: a chief holding the record scope must be able to read the patient's own notes.

**I could not reproduce it.** Full account below, plus the test this PR actually adds.

## Reproduction attempt

Fresh worktree, `make reset-db`, dev server on port 8140, `NURA_BASE_URL=http://127.0.0.1:8140 make checkpoint N=13`:

- **At `e76c223`** (the SHA the other builder confirmed the failure on, on a throwaway worktree): checkpoint 13 passed, 3/3 clean runs. Step 4 ("Pa let Mei in to his private notes as well ... Mei reads the note") succeeded every time.
- **At `main`'s tip when I started (`20d32dc`)**: checkpoint 13 passed, 5/5 clean runs.
- **At `main`'s tip once it moved again mid-session (`a3a0a4b`, #166 landed)**: checkpoint 13, 2 and 16 all pass.
- Full backend suite (`make test`) is green at each point: 2261 passed, 1 skipped, 0 failed on `a3a0a4b`.
- `make lint`, `make plain-words`, `make language` are all clean (0 failures) on the branch tip.

## Bisect

I looked at every commit between `e76c223` and `20d32dc` for anything touching keys/consent/audit/family/notes: only `e9181c8` (#170, "refusals on the trail") touches that area, in `backend/app/audit/strings.py` and `backend/app/family/trail.py`. That diff rewords the delivery engine's own daily read-count line and the "stewardship" trail phrase — it doesn't touch `Scope.NOTES` refusal wording or the notes read path at all. Nothing else in #169/#147/#180/#179 touches `app/notes/`, `app/keys/scopes.py`, or the notes branch of `app/audit/access.py`.

My best guess: the original report came from a contaminated throwaway worktree (a stale dev server still serving older code on the target port, or a `dev.db` that wasn't actually reset) rather than a code regression. I can't verify that remotely, so I'm stating it as a guess, not a finding.

## What I did instead

No application code changed — I found no cause to fix. What I did add: a **pytest-level** regression test, because the only thing that ever walked this exact scenario was the checkpoint script, which is dev-only and **never runs in CI** (checked `.github/workflows/`). Nothing in the pipeline was actually guarding this flow. Separately, the existing pytest coverage in `test_family_privacy.py` exercised a chief and a caregiver ("Kit") against the notes scope, but never a `VIEWER` or `HELPER` key specifically — which is exactly what the acceptance criteria for this task asked to pin.

New test: `backend/tests/test_family_privacy.py::test_chief_reads_his_notes_a_viewer_and_a_helper_are_refused_on_his_trail`

- Mei (chief, `ROLE_SCOPES[KeyRole.CHIEF]` = every scope) reads Pa's note.
- A newly granted viewer key (Priya) is refused, `OutOfScope`.
- The household's existing helper key (Siti) is refused, `OutOfScope`.
- Both refusals are on Pa's own trail, in his words: `"{who} asked to see your private notes on {day}." / "Only you can."`.

## Checks run

- `make checkpoint N=13`, `N=2`, `N=16` — all pass, against the branch tip.
- `make test` — 2261 passed, 1 skipped, 0 failed.
- `make lint` — clean.
- `make plain-words` — 0 failures.
- `make language` — 0 failures.

## Reviewer reports

Ran `.claude/agents/clinical-safety-reviewer.md` and `.claude/agents/plain-words-reviewer.md` by hand over the diff (the `review` CI check is out of API credits, per instructions, and is ignored here).

### clinical-safety-reviewer

> Scope of diff: test-only, one new test function plus new imports. No application code under `backend/app/` changed. Per checks 1-8, most are N/A since nothing about behavior could have regressed or been introduced here.
>
> 1. Every profile read goes through the keys module with a key context — **PASS**. Every read in the new test goes through `list_notes(sg, context=...)` with a `KeyContext`.
> 2-6. **N/A** — no Fact, Card, medicine, drug data, red-flag or high-risk content touched.
> 7. Health data stays in region — **PASS**. `register_person(sg, region=Region.SG, ...)`.
> 8. Consent recorded at claim/per key, refusals visible in audit — **PASS**. `agree_to_family_sharing` (consent) runs before `grant_key`; both refusals asserted in `read_audit` with `OutOfScope`, and separately on the plain-words trail.
>
> Required changes: **none**.

### plain-words-reviewer

> Both literal strings in the diff are pre-existing, already-reviewed catalogue copy — traced to `app/audit/strings.py:157` (the `ONLY_YOU` family), `app/consent/texts.py:45` (`Scope.NOTES` → "your private notes"), and `app/medicines/strings.py` (`say_date()`) — correctly reproduced by the new test, not new copy.
>
> - "Priya asked to see your private notes on Monday 14 September." — **PASS** (matches the glossary's "private notes"; day+date format is the rule's own example).
> - "Only you can." — **PASS** (matches the glossary's approved "Only you can" phrase).
>
> Nothing to rewrite. No new patient-facing copy was introduced.

---


ashfrnndz21:
**Operator merge note.**

Merging. The `risk:high` label came from the area (keys/consent/notes), not from the diff: the diff is one file, `backend/tests/test_family_privacy.py`, and adds one test. No application code changes, so there is no behaviour here that could regress — that is why this one does not wait on an independent review agent the way a risk:high code change does.

On the reported regression itself I accept the builder's account: checkpoint 13 passed on the SHA the failure was first seen on (3/3), on main's tip when work started (5/5), and on main's tip after #166 landed, with the full backend suite green at each point. The bisect over every commit between those points found nothing touching `Scope.NOTES`, the notes read path, or `app/keys/scopes.py`. A contaminated throwaway worktree is the most likely explanation, and the builder was right to state that as a guess rather than a finding.

The thing worth keeping is the reason this was invisible: the only thing that ever walked this scenario was the checkpoint script, and **checkpoints do not run in CI**. So a chief losing his notes would have reached a release without a single red test. The new pytest covers a chief reading, a viewer refused and a helper refused, all three on Pa's own trail in his words. That gap — checkpoints being dev-only — is worth its own issue, and I'll open one.

ashfrnndz21:
**Correction to my merge note on this PR.**

When I merged this, I accepted the builder's guess that the reported failure of checkpoint 13 step 4, "Mei reads Pa's notes as his chief", came from a contaminated throwaway worktree: a stale server or a database that had not been reset. The builder was careful to call it a guess. I wrote it into the record more confidently than it deserved.

**That guess was wrong.** Building the checkpoint runner for #192 (PR #218), a builder ran the 18 runnable checkpoints with and without a frozen clock. **Ten of them fail only when the clock is frozen, deterministically**, and checkpoint 13 is one of them, failing on exactly that step.

The cause: `backend/app/audit/trail.py:147` orders the trail by `AuditEntry.at.desc()` with no tie-breaker. The primary key is a random UUID4, so it cannot break a tie in time order. When several events share one instant, "newest first" is arbitrary, and a step that reads the latest entry gets whichever one the database happens to return. The frozen clock makes ties constant. Two writes in one request from a single clock read make them possible in production too.

So the original report was a real, reproducible bug that looked random. It went unreproduced here because runs without a frozen clock rarely produce ties.

It is not confined to the trail. Sixteen queries pick "the newest" by timestamp alone, including ones that decide current state: which delivery settings apply, which login challenge is valid, which red flag is current, which "which tablet?" question is open. That fix is being built now, `risk:high`.

The test this PR added is still worth having. What was wrong was the explanation, not the test.

PR #191, opened 2026-09-16
A red word goes first, whatever else the app was about to do

ashfrnndz21:
## The cause

`day.spec.ts:177` asserts a red word's `POST .../not-feeling-well` is the *first* request after he speaks. In CI (run 35065730744) `GET .../feed` sometimes went first instead.

The client's request queue (`web/src/api/client.ts`) already sent an `urgent: true` call ahead of every call still *waiting*, but the code — and its own comment — said "behind only the one already on the wire." Today's own page load (`refresh()` in `web/src/screens/Today.tsx`) reads profile/state/medicines/doses/proud/feed sequentially in the background, and is never cancelled when he navigates to Not Well, Symptoms or taps the feeling cloud (which is inline on Today, so it doesn't even unmount). Under CI's slower timing, that background `GET .../feed` was sometimes still on the wire at the exact moment the urgent POST was enqueued, and the old queue let it finish first — a genuine race, not a fixed ordering bug, which is why it reproduced in CI but not reliably on a fast laptop.

## The fix

`web/src/api/client.ts`: the queue now tracks the job presently on the wire. An urgent call aborts it if it is a non-urgent, **abortable** (GET-only) read, and is sent in its place at once — the aborted call's caller sees `Unreachable`, exactly like a lost network, which the app already handles gracefully everywhere (kept/cached page stands in, nothing is lost). A **write** already on the wire (a held tap replaying, a nudge answered) is deliberately left alone — it may already be at the backend, so aborting-and-retrying could double it — and the urgent call is sent the moment it clears, never later. This holds on every route into a red word: the voice path, the typed path, the symptom log, and the feeling cloud (all four `urgent: true` call sites in `web/src/api/nura.ts`), because the fix is in the shared queue itself, not per-screen.

`web/tests/unit/queue.test.ts` pinned the old "behind only the one already on the wire" behaviour as correct; rewrote it to assert the new guarantee (read aborted and preempted; write left to finish; urgent call still gives up at its own deadline if truly stuck).

`web/tests/e2e/day.spec.ts` gets one new regression test: it holds a real `GET .../feed` open with `page.route` across the whole red-word voice flow, so the race is deterministic instead of load-dependent, and asserts the POST is still first and the flag still lands.

Offline and reopen: unaffected — a red word is never held in the offline queue (`offline/queue.ts`: "a red feeling word is refused here"; the not-feeling-well path shows the backend's offline card directly on a lost network), and `Today.tsx`'s `load()` already replays held taps before its own refresh, so nothing here changes that order. Ran `offline.spec.ts` and the offline/hung-request tests in `day.spec.ts` — all green.

## Verification

- Reproduced on a fresh worktree from main's SHA `20d32dc`, `make build-web`, `npx playwright test tests/e2e/day.spec.ts:177` — passed in isolation locally (the race needs CI-level load to show up reliably), confirmed the actual cause via the CI log and by reading `Today.tsx`'s `refresh()` and the request queue.
- `day.spec.ts` (whole file, including the new test) — 3 repeat runs, all green.
- `offline.spec.ts` — green.
- Full web unit suite (`vitest`) — 300 passed.
- Full web e2e suite — 111 passed.
- Backend `make test` — 2271 passed, 1 skipped, 0 failed.

### clinical-safety-reviewer

Scope: transport-only change to the request queue (`web/src/api/client.ts`) plus the two test files. No backend, medicine, drug-data, consent, audit, or region file touched.

1. Every profile read goes through the keys module with a key context — **PASS**. No new call sites; the four `urgent: true` sites in `nura.ts` are unchanged and still route through `api()`, which still attaches `Authorization: Bearer`.
2. Every new Fact carries provenance; every Card records its State — **N/A**. No Fact/Card construction in this diff.
3. No output starts, stops or changes a medicine — **PASS**. `apiUpload` (writes) is explicitly hard-coded non-abortable, so no in-flight write (e.g. a dose `taken` POST) can be aborted/duplicated.
4. Drug identification, interactions and dosing come from the licensed data client only — **N/A**.
5. Red-flag paths escalate before any ranking or caps — **PASS, strengthened**. The urgent lane now also aborts a non-urgent, abortable job already on the wire, not just calls still waiting, closing exactly the CI race. `URGENT_DEADLINE_MS` abort-at-deadline logic is unchanged — an urgent call still gives up as `Unreachable` at 10s and is never sent late.
6. High-risk drugs require a label photo before a dose is saved — **N/A**.
7. Health data stays in region; no health data reaches analytics — **PASS** (no region/analytics code anywhere in the diff).
8. Consent is recorded at claim and per key; refusals are visible in the audit — **PASS** (no consent/audit code touched).

Additional: confirmed `Today.tsx`'s `refresh()`/`load()` already tolerates a background read failing (`Unreachable` → `setUnreached`, never `forget()`), and `feed/store.ts` and `emergencyCache.ts` both already branch on `Unreachable` to keep what's cached rather than clear it — so aborting a background read cannot produce silent data loss or a stale page shown as authoritative.

Required changes: none safety-relevant. (One nit — a stray spec-ID citation in a comment — already fixed before this PR was opened.)

### plain-words-reviewer

No patient-facing strings in this diff. `web/src/strings/{en,ms,zh}.ts` are untouched. Everything added is internal code/JSDoc comments, the internal error class name `Unreachable` (used only in `instanceof` checks), Playwright/Vitest test titles, `getByTestId` selectors, and data discriminants like `card.kind === "red_flag"` — none of it rendered to Pa or Mei. Verdict: nothing here needs plain-words review.


ashfrnndz21:
**Operator note — independent review, then merge.**

`risk:high` on the red-flag path does not merge on the builder's own review here. I sent this to an independent reviewer with two specific doubts, because they were the two ways this change could hurt someone:

**1. Could it abort a write?** `api()` derives `abortable` from the method, but `apiBlob()` and `apiText()` hard-code `true`. If any caller passed a non-GET, a red word could cancel a write — the exact thing the PR promises never happens.

Answered, and better than I expected: `sendBlob` (`client.ts:199`) and `apiText`'s inner fetch (`client.ts:225`) **hard-code `method: "GET"` in the `fetch()` call itself**, ignoring `call.method`. So it is not merely that no current caller passes a method — a future one could not produce a non-GET on the wire through those functions even if it tried. `apiUpload`/`sendBytes` is symmetric: always `abortable: false`, always hard-coded `POST`. The five real callers (`consentRecord`, `emergencyCardPage`, `feedVoice`, `clip`, `storyVoice`) were each checked.

**2. Does an aborted read show Pa an error?** An error card appearing *because he tapped a red word* would be worse than the delay this removes. All three named call sites were traced: Today's refresh maps `Unreachable` to `setUnreached("network")` and keeps the stale model on screen; the feed's prefetch sets `offline` and leaves `entries` untouched; the restore after sign-in returns `"stay"` and is invisible. The unnamed ones too — `Playback.warm` drops it and falls back on the next `hear()`, `StoryVoice.note` falls back to the phone's voice, and the two that surface it render the friendly `Notice`. No dead ends.

Also checked: the `current`-pointer window (aborting an already-settled controller is a no-op, and `pump()`'s `finally` always clears it); urgent ordering under a burst; and whether the unit test is tautological — it is not. The reviewer disabled the `current.control.abort()` line and `queue.test.ts` failed immediately on the right assertion.

**Two honest limits**, recorded rather than papered over: the new `day.spec.ts` e2e was not executed against a live backend, so "it fails before the fix" is unverified by execution (the mechanism is the same one the unit test proved); and there is no end-to-end test of an urgent call arriving *mid held-tap-replay* — the unit test's write-not-aborted case stands in for it, but not through the real offline-replay path. Both are coverage gaps, not defects.

**Verdict: MERGE**, no blocking items.

Separately, the review turned up a pre-existing defect this PR neither causes nor worsens: `sendBlob` skips `fetchWithin`, so `feedVoice`, `clip` and `storyVoice` have no `CALL_DEADLINE_MS` ceiling and a stalled audio fetch hangs for ever — Pa taps *Hear* and nothing ever tells him, because the existing fallbacks trigger on failure and a hang is not a failure. Filed as #193.

Issue #196, opened 2026-09-16
The Postgres suite is at the 30-minute job limit and gets cancelled at random


ashfrnndz21:
## What is wrong

`backend-postgres` runs the whole backend suite against a real Postgres. It is the only place a Postgres runs at all, and it is a required check.

It now takes **27–30 minutes** and has started being cancelled at the 30-minute job limit at random. On #191 it was cancelled at exactly 30:03; on #190 it finished at 27m20s. A cancelled job reports as a failed check, which looks identical to a real test failure — I have now twice pulled a job log expecting a regression and found only `##[error]The operation was canceled.`

I have raised the limit to 60 minutes as a stopgap. That stops the false reds but makes every PR slower to merge, and the suite will keep growing.

## Why it is slow, and the thing that makes it fixable

`backend/tests/conftest.py` already gives **each test its own schema** on the shared server through `NURA_TEST_DATABASE_URL`, and drops it afterwards. That is the hard part of parallelising a database suite, and it is already done. The tests do not share state; they are simply running one after another.

## What to build

Run the suite in parallel with `pytest-xdist` (`-n auto`), and make the schema-per-test machinery worker-safe:

- Confirm each xdist worker gets schema names that cannot collide (the worker id belongs in the name).
- Check the migration steps that run *before* the suite (`alembic upgrade`, the down-to-nothing-and-up-again step, `alembic check`) — those are sequential and must stay sequential, and must not race the workers' schema creation.
- Watch for tests that quietly depend on running in a particular order, or on being alone on the server. Any that surface are real findings: write them up rather than pinning them to a single worker to make the run go green.
- The frozen clock (`app/clock.py`, `NURA_FROZEN_CLOCK`) is per-process, so check nothing depends on a clock shared across tests.

Target: the same suite, the same assertions, under ten minutes. Then bring the 60-minute limit back down — a limit well above the real runtime hides the next slowdown just as effectively as one below it hides nothing.

## Do not

Do not make the suite faster by skipping tests on Postgres, marking them slow and excluding them, or splitting the job so only part of the suite is required. The point of this job is that the whole suite runs against the database the deployment actually uses.

Issue #192, opened 2026-09-16
The checkpoints never run in CI, so an acceptance walk can break with every test green


ashfrnndz21:
## What is wrong

`backend/scripts/checkpoint.py` and everything under `backend/scripts/checkpoints/` walk the acceptance criteria end to end against a running dev server. They are the closest thing this repo has to "does the product actually work".

Nothing in `.github/workflows/` runs them. They are dev-only, run by hand.

This surfaced on #190: checkpoint 13 step 4 ("Mei reads Pa's notes as his chief") was reported failing on a clean checkout. It turned out not to be a real regression — but if it had been, **every CI check would have stayed green while a chief silently lost the ability to read the notes he holds the key for**. The only pytest coverage in that area exercised a chief and a caregiver, never a viewer or a helper key specifically. #190 pins that one flow in pytest. The other checkpoints have the same hole.

## What to build

A CI job that runs the checkpoints. The shape that probably works:

- Boot the app the way `make dev` does, on a throwaway database, with the frozen clock (`NURA_FROZEN_CLOCK`) and the fixture providers the dev target already sets.
- Run every checkpoint marked **ready** in `docs/checkpoints.md`, not a hand-picked subset — the table is the list, and a new checkpoint should join the job by being marked ready, with no workflow edit.
- Fail the job on the first failing step, and print the step's own words, not a stack trace: the checkpoints are written in the product's language and that is what makes them readable when they break.
- It will be slow. Run it on pull requests against `main` and keep it out of the required set at first (strict=false, like `image`), so we learn its flake profile before it can block a merge.

## Watch out

- The checkpoints assume a clean database. A shared or reused one is exactly the contamination that produced the false report on #190, so the job must create and destroy its own.
- Each checkpoint must pick a free port rather than a fixed one, or parallel jobs will serve each other's code — the same failure mode again.
- Do not let the job paper over a failure by retrying. A flaky checkpoint is a real finding about the product.

PR #199, opened 2026-09-16
E22-03: a card's spoken twin is said when the card is made; the Voice port gets a conformance suite

### LINKED ISSUE #88: E22-03 Voice scripts: every card rendered as a spoken script with pauses; pre-rendered
**Epic** E22 Plain words and voice
**Persona** System · **Tier** T1 · **Priority** P1 · **Size** M

**Story**
Voice scripts: every card rendered as a spoken script with pauses; pre-rendered audio per language; synthesiser fallback

**Acceptance**
Every patient card has audio within a minute of creation

**Depends on**
E22-02

**Spec** see docs/00-MASTER-BUILD-SPEC.md section 0 for the document that covers this epic.

ashfrnndz21:
Closes #88.

`docs/parity.md` said: *"Not met: 'audio within a minute of creation'. Audio is rendered on first play, not when a card is made (`twin.py`: 'rendered to audio on request'), and the only voice is the fixture's silence; there is no speech provider in the region."*

Two gaps, addressed separately.

## 1. Pre-render at creation

`_say_ahead` now runs as each card is made, through the one voice port into the region's store, under the digest of the card's voice script — the same lines, language and boundary the twin route says, so the same key, and the first play is a read.

The care is in the failure path. Each item renders in **its own savepoint**, so a voice that is down never costs him the card `create_item` already wrote. But the failure is not swallowed into a log line either: it is written to the trail under `feed_item_voice`, on the card's own scope, as a refused write. That target name is deliberately not `feed_item` — a render failure has to be tellable apart from a refusal to write the card itself. The gap between "the card exists" and "its twin is ready" is now visible to whoever reads the trail.

A card with no voice in its language yet, or too long to say, is not said ahead, and the route answers for it exactly as it always has.

## 2. The `Voice` port, made real

`Voice.speak` now takes the **whole script** rather than a flat string: one segment per line, each carrying the pause to leave after it, the boundary's longer than the rest. A provider that can control its own pauses (SSML `<break>`, or whatever its API calls it) reads `segment.pause_ms` off each segment; one that cannot still says the lines in order, and the fixture's duration still counts every pause — so the thirty-second rule holds without a speech provider.

`backend/tests/voice_conformance.py` is the suite **any** adapter must pass: the right language, the duration bound, the pauses the script asks for, Hokkien and Tamil refused until T2, a note past the limit refused. The fixture is proven against it here, so a regional provider drops in with no caller change.

**The honest remaining gap, stated in `docs/parity.md` rather than glossed:** the provider itself. There is still no licensed speech provider in-region, and this PR adds none and no credential. What it does is make the seam real enough that adding one is an adapter, not a rewrite.

## Operator amendments before opening

I finished this one off myself — the builder had the work done and gated but stalled. Two changes on top of its branch:

- **A plain-words failure its own reviewer caught.** The zh line for "It is in your blood pressure book." read `它记录在您的血压手册里。` — `docs/plain-words.md` §6 fixes the noun as **血压本**, not 血压手册 ("handbook"), and 记录在 ("is recorded in") drifts to a clinical register the English ("is in") and Malay ("ada dalam") lines do not use. Now `它在您的血压本里。`
- **A docstring that was wrong in a way that mattered.** `_say_ahead` claimed *"on main this whole function is a no-op (`engine.voice`/`engine.store` are None) until a deployment names both"* — which would have told a future reader that pre-rendering is not live. The fields are optional on the `Engine` so a test may omit them, but `_engine` in `app.channels.api.feed` always passes `providers.voice` (the fixture voice by default) and `providers.object_store` (required). It runs on every refresh, dev included. Corrected.

## Gates

Per the operator's standing rule (this Mac runs several builders at once; concurrent full suites were being killed by contention), the **full suite is deferred to CI** — `backend` and `backend-postgres` are the gate for it. Run locally:

- `pytest tests/test_voice.py tests/test_voice_ahead.py tests/test_voice_conformance.py tests/test_voice_script.py tests/test_feed.py` — **70 passed**
- `ruff check .` — All checks passed
- `mypy app` — no issues in 259 source files
- `make plain-words` — 7112 strings in 32 files, **0 failures**
- `make language` — 7245 strings, 2441 keys in 30 catalogues, **0 failures**

The `review` check is out of API credits and always fails; it is not a gate. If `backend-postgres` goes red, check the log for a cancellation before debugging — see #197 and #196.


ashfrnndz21:
Checkpoints 20 and 23 run on port 8240, against the branch as it stands now (commit `a911a4d`, includes both amendments).

**Checkpoint 20: passed.**
```
✓ "Your tablets today" played as its spoken twin (GET …/feed/{item}/voice): audio/wav, 92044 bytes, 11.5 seconds, cache hit — the fixture voice is silence as long as the words take to say
```
`cache hit` on the very first play — a card created earlier in the run had its audio already waiting, not rendered on request. Every other step in the checkpoint also passed.

**Checkpoint 23: passed.**
```
✓ Pa (132/81, Mei holds a key): every card on his feed carries its voice script, computed from its verified voice lines — the reading card is said:
    Your blood pressure today was one hundred and thirty-two over eighty-one.  ⟨500 ms⟩
    It is in your blood pressure book.  ⟨500 ms⟩
    Mei can see it too.  ⟨500 ms⟩
✓ the learning card pauses longer before the boundary it ends on:
    This comes from HealthHub.  ⟨1200 ms⟩
    Nura explains one thing in simple words.  ⟨500 ms⟩
    This is not a doctor's advice.  ⟨500 ms⟩
    Ask your doctor.  ⟨500 ms⟩
✓ Kit's card in Chinese: "您今天的血压是132比81。" is said "您今天的血压是一百三十二比八十一。"
✓ Aminah's card in Malay: "Tekanan darah anda hari ini 132 atas 81." is said "Tekanan darah anda hari ini seratus tiga puluh dua atas lapan puluh satu."
```
All three T1 languages carry a correct voice script; the boundary line's 1200ms pause is distinct from the ordinary 500ms line pause, in each direction the checkpoint walks. Every other step (the pharmacist's queue, the rewrite, the source approval) also passed.

`checkpoint 20 passed` and `checkpoint 23 passed` were both the last line of their runs — full logs available if useful, but nothing failed.

ashfrnndz21:
**Operator note — independent review: BLOCK, now fixed. Pushed `560bc7c`.**

I sent this for independent review even though it is `risk:medium`, because I finished the PR myself and nobody had looked at it with fresh eyes. That was the right call: it found a real defect, and a worse one than I had guessed at.

**The defect.** `_say_ahead`'s except-handler wrote the trail line inside its own `nested_unit_of_work` — but that block had no enclosing try/except of its own, unlike the render above it. Writing the trail line is itself a write. If `record()`'s flush failed for any non-refusal database reason — a dropped connection, a statement timeout, a serialization conflict — the exception left `_say_ahead`, left `refresh`, and reached the request-scoped `unit_of_work` in `app/channels/api/deps.py`, whose own docstring says: *"Any other failure rolls the whole request back."*

So the compound case — the voice is down **and** the trail write is unlucky — would have unmade **every card that refresh created**, not just the card whose voice failed, and dropped the trail lines earlier items in the same loop had already earned. That is exactly the failure the per-item savepoints were added to prevent, and the docstring promised it could not happen. It was also untested: the only injected failure in the suite was inside `voiced()`, never inside `record()`.

**The fix** follows the sibling pattern already in the same file — `_sample()` wraps its whole body so its bookkeeping never reaches the caller. The trail line now degrades to a log line, and the cards stand.

**The test injects the failure where it had never been injected** — in `record()` — and asserts the card row survives and that no trail line was written, which is the cost we accept. I mutation-checked it rather than trusting that it passes: with the guard removed it fails with the `OperationalError` escaping `refresh`, so it holds the actual bug and is not asserting the implementation back to itself.

**What the review verified and I am not re-doing:** the first savepoint is correct (and `voiced()` touches no session at all, so no aborted-transaction state is possible there); `except Exception` does not swallow `CancelledError` (a `BaseException` since 3.8) and cannot see a scope violation, since `voiced()` takes no `KeyContext`; `ck_audit_entry_actor_or_system` is satisfied on Postgres as well as SQLite because `channel = 'system'` alone discharges the OR; `refused_because` carrying only a class name is deliberate repo convention (*"the name of the refusal, never what was held back"*); and both production `Engine(...)` construction sites pass a voice and a store, so pre-rendering has no silent gap.

**Two things I am not blocking on, recorded so they are not lost:**

1. **The "within a minute" acceptance is met by architecture, not by measurement.** The test asserts `clock.now() == made_at` on a frozen clock, which is structurally true of any synchronous path — a real provider taking five minutes would still pass it. What it genuinely proves is that pre-rendering happens inside `refresh()` before the response returns, which is the right shape. Checkpoint 20 backs it end to end by showing `cache hit` on the first play. But the bound itself is untested and, with a frozen clock, close to untestable.
2. **`Channel.SYSTEM` alongside a real `actor_person_id`.** The trigger engine's convention is to convert context with `as_the_system()` first; `_say_ahead` forces `SYSTEM` while passing the request's own context, so the row carries both. There is direct precedent — `app/safety/red_flags.py::_system_read` does the same deliberately — so it is a second use of an existing pattern rather than a new inconsistency. Worth a look when someone next touches audit conventions.

**Gates after the fix:** `pytest tests/test_voice_ahead.py tests/test_voice.py tests/test_feed.py` — 28 passed; ruff clean; mypy clean on 259 files. Checkpoints 20 and 23 passed on the prior commit, with checkpoint 20 showing `cache hit` on the first play — the acceptance demonstrated end to end rather than inferred.

ashfrnndz21:
**Operator merge note.**

Merging. Independently reviewed, blocked, fixed, and re-verified.

**What it closes.** A card's spoken twin is now said and kept when the card is *made*, not on its first play, under the digest of its voice script — so the first play is a read. And the `Voice` port is a real seam: `speak` takes the whole script (one segment per line, each carrying the pause after it, the boundary's longer than the rest), with `tests/voice_conformance.py` as the suite any adapter must pass — right language, the duration bound, the script's pauses, Hokkien and Tamil refused until T2, a note past the limit refused. The fixture is proven against it, so a regional provider drops in with no caller change.

Checkpoint 20 showing **`cache hit` on the first play** is the acceptance demonstrated end to end rather than inferred; if audio were still made on demand it would be a miss. Checkpoint 23 confirms the scripts in all three languages with the two distinct pause lengths. Neither checkpoint runs in CI (#192), which is why they were run by hand.

**The blocking defect the review found**, and the reason this PR is better than it was: the `record()` call in `_say_ahead`'s except-handler sat in its own savepoint but had **no enclosing try/except**. Writing the trail line is itself a write, and if that flush failed for any non-refusal database reason it left `_say_ahead`, left `refresh`, and reached the request's own unit of work — whose documented contract is that any failure that is not a refusal rolls the whole request back. So the compound case, voice down *and* one unlucky write, would have unmade **every card that refresh created**, not just the one whose voice failed, and dropped the trail lines earlier items in the loop had already earned. Exactly the failure the per-item savepoints were added to prevent, and the docstring promised it could not happen.

It was untested, too — the only injected failure in the suite was inside `voiced()`, never inside `record()`. The new test injects it where it had never been injected, and I mutation-checked it rather than trusting a green: with the guard removed it fails with the `OperationalError` escaping `refresh`. It holds the actual bug.

**Two things recorded rather than glossed.** The "within a minute" acceptance is met by architecture, not measurement — the frozen-clock assertion is structurally true of any synchronous path, and a real provider taking five minutes would still pass it. What it genuinely proves is that rendering happens inside `refresh()` before the response returns; checkpoint 20's cache hit is the stronger evidence. And `Channel.SYSTEM` is written alongside a real actor id, which differs from the trigger engine's `as_the_system()` convention but has direct precedent in `app/safety/red_flags.py::_system_read`.

**The honest gap, stated in `docs/parity.md` rather than glossed:** there is still no licensed speech provider in-region, so the fixture's silence is the only voice anywhere runs. This PR adds none and no credential. What it does is make adding one an adapter rather than a rewrite.

Also merged: a plain-words fix its own reviewer caught — the zh line for "It is in your blood pressure book" said 血压手册 where the glossary fixes 血压本, and 记录在 ("is recorded in") drifted to a clinical register the English and Malay lines avoid.

PR #202, opened 2026-09-16
E00-06 + E04-03: licensed drug registry grown to 233 products; supplements and TCM screened for interactions

ashfrnndz21:
Closes #6
Closes #28

## The two acceptance lines

**#6 E00-06** — *"Brand to generic resolution for a sample of 200 MY/SG products above 95%"*
Met, and measured, not claimed. `backend/tests/fixtures/drugs/registry.json` is grown from 17 to 233 real MY/SG community-pharmacy and chronic-disease products (50 generics: cardiovascular, diabetes, lipids, analgesia, gastro, respiratory, plus supplements and TCM), with plausible NPRA/HSA-shaped registration numbers. `backend/tests/drug_resolution_accuracy.py` is a harness in the style of the E02 capture-accuracy harness (`tests/paper_accuracy.py`): it resolves every product's own brand+strength back to its own generic through `DrugRegistry.identify()` and scores it. `backend/tests/test_drug_resolution_accuracy.py` holds it above 95% and fails loud with every miss printed. Today: **100.0% over 233 products, 0 misses.**

**#28 E04-03** — *"New line screened before save; severity and the two drugs named"*, story title *"…including supplements and TCM"*
Met, including supplements and TCM. A product now carries its own `ProductKind` (prescription, supplement, TCM) from the registry, and is screened for interactions the same way regardless of kind. New pairs: warfarin with ginkgo biloba, dong quai, danshen, high-dose fish oil, vitamin K and ginseng; St John's wort with an SSRI (sertraline, serotonin risk) and with a DOAC (apixaban, weakens it); an NSAID (ibuprofen) with warfarin/apixaban/clopidogrel; potassium with an ACE inhibitor/an ARB/spironolactone; calcium or iron with levothyroxine or a quinolone. `POST /profiles/{id}/medicines/draft` flags before save, severity and both products named, in English, Malay and Chinese — the route's `language` query param was previously ignored (always English); it now resolves to the profile's own, matching every other medicines route. `backend/tests/test_supplement_tcm_interactions.py` is the acceptance line as tests.

## The port and the pharmacist review queue (E00-06 → E04-03's seam)

[...]

PR #195, opened 2026-09-16
E11-01: a heard WhatsApp voice note classified like a typed message; a whole day proven end to end

ashfrnndz21:
Closes #41
Refs #78

## What I found before building

`docs/parity.md` (audited on an older `main`) said #41 (E11-01) was missing the WhatsApp
family group thread and voice notes from Pa, and that no test walked a whole day (#78,
E19-03, shares that acceptance line). Reading the current code first: **the family group
thread and voice notes from Pa were already built and tested on `main`**, by prior PRs
#146, #168 and #180 (`app/channels/whatsapp/group.py`, `app/family/thread.py`, the voice
handling in `app/channels/whatsapp/inbound.py`, `tests/test_whatsapp_family_voice.py` — 16
tests, all green on `main` before this PR). `docs/parity.md` had gone stale. There was also
already a `tests/test_whatsapp_day.py` proving most of E19-03's "a full day without opening
the app" acceptance line.

So this PR does **not** rebuild #1 or #2 of the story brief. It closes the one real gap I
found, and fills out the whole-day test to the brief's fuller scenario.

## What was actually missing, and what this PR does

**The gap:** a heard WhatsApp voice note always became a generic "kept as his own note" —
even when its transcript said "Taken" or "sudah makan ubat" (already took the medicine). It
was never classified through the same door a typed message is, so Pa's own voice could
never write the Taken tap or answer an open check-in — only typing could.

- `backend/app/channels/whatsapp/inbound.py`: in `_dispatch`, a heard, non-group voice note
  (past the red-flag check, which is unchanged and still runs first) is now classified via
  the existing `classifier.classify()` exactly as a typed message is: `Kind.TAKEN` routes to
  `_taken()`, `Kind.ANSWER` routes to `_answer()` — the same functions the typed "Taken" and
  "OK" paths already use and already test. Everything else (a document, a health event,
  coordination, anything else) still falls through to `_voice_note()`, kept as his own note,
  unread. An unheard/mumbled note is unaffected: its transcript text stays `None`, so it can
  never match `TAKEN`/`ANSWER` — it is never guessed at.
- Two new voice fixtures (`tests/fixtures/voice/*.json` + `media.json` entries):
  `pa-voice-taken` ("sudah makan ubat") and `pa-voice-ok` ("OK"), plus two new tests in
  `tests/test_whatsapp_family_voice.py` proving the Taken tap and a check-in answer both
  write from his voice the same way typing does.
[...]

PR #209, opened 2026-09-16
#206: a drug match is scored, not just found — a confidence floor before a line is identified

### LINKED ISSUE #206: NotIdentified means 'the registry missed', not 'the label is bad' — and gets weaker as the registry gets real
## What was found

`test_the_label_must_name_a_medicine_and_say_the_dose_one_way` broke on #202, and the cause is more interesting than the break.

Growing the fixture registry from 17 to 233 products made `ibuprofen 200mg` — previously absent — match a real product. The refusal stopped firing, and a label the test says must be refused was accepted with a 200.

Investigating what the refusal actually keys on gave the finding: **`NotIdentified` is purely a registry-lookup miss.** It has no opinion on whether the label is well-formed. The test's name promises that a label "must name a medicine and say the dose one way"; what the code enforces is "this string matched nothing in the registry we happen to have".

The immediate break is fixed in #202 by switching the input to `digoxin`, a generic deliberately and permanently excluded from the fixture. That is a fair repair and it keeps the test honest today. It does not address the shape of the problem.

## Why this matters as the data gets more real

The fixture has 233 products. A licensed NPRA/HSA feed has tens of thousands, including near-homonyms, brand families, and strengths across many forms. The larger and more real the registry, the more likely it is that a malformed, garbled or half-read label **matches something** — and therefore the *less* often this refusal fires. The check gets weaker exactly as the data gets better, which is the wrong direction for a safety check to move.

Worth being precise about what is and is not at risk: there is a genuinely content-based validation earlier in the same route that returns `422` for a label that is not well-formed, and that is unaffected. The gap is the space between "this label is syntactically plausible" and "this is really the medicine in his hand" — which today is decided entirely by whether a lookup hit.

It is also not unique to our fixture. A real registry adapter behind the same port inherits the property, because the property is in the check, not the data.

## What to decide

This needs a decision before a patch, and it may be that the current behaviour is right and only the naming is wrong:

- **If the refusal is meant to be about the label's content**, then it needs a content test of its own — does the text name something in a medicine-shaped way, does it state a dose unambiguously and once — independent of whether the registry knows the product. A miss in the registry then becomes a *different* outcome ("Nura does not know this one") rather than the same one.
- **If the refusal is only ever meant to be a lookup miss**, then rename it and the test so nobody reads a guarantee into it that is not there — and make sure the high-risk-drug path that requires a label photo before a dose is saved is not leaning on this check for something it does not provide.

The second question is the one I actually care about: **what else treats `NotIdentified` as though it meant "this label is not trustworthy"?** That should be audited before either fix.

## Credit

Found by the builder on #202 while fixing a test it would have been much easier to quietly repair. It reported the underlying property plainly instead, which is why this issue exists.

ashfrnndz21:
Closes #206

## The decision this closes

From #206, the operator's chosen fix:

> "when it recognizes the med, it should give it a confidence score" ... The problem was never that identification is registry-based — it is that a match is binary, so a garbled label that happens to collide with a real product is indistinguishable from a clean exact one. A confidence score makes those two different things again, and it *improves* as the registry grows rather than degrading ... It also makes this the third use of one rule already in the codebase ... `CONFIDENCE_THRESHOLD = 0.8` ... **Nura does not act on a recognition it is not sure of.**

Built on a fresh branch off current main (not on #202, which stays a separate, already-green PR).

## What changed

- `DrugMatch` (`app/drugs/registry.py`) carries a `confidence: float`. `FixtureRegistry.identify()` (`app/drugs/fixture.py`) scores every candidate: **1.0** for a registration number, or a name whose strength (and form, where given) both belong to the matched product; **0.4** when a given strength does not belong to it; **0.5** when a given form does not; **0.9** for a bare name match with nothing to check it against; **0.5** — below the floor — for a bare name match on a **high-risk** product (see below). Results come back best-first.
- `app/medicines/service.py::_one_product` — the one place `plan()` turns a registry answer into a line — refuses (`NotIdentified`) anything below `CONFIDENCE_THRESHOLD` (imported from `app.ingestion.models`, not a new number). This is the same floor a document field extracted from a photo and a WhatsApp voice transcript (#195) are already held to.
- The confidence is recorded on the line beside the provenance already there (`medication_line.registry_confidence`, migration `0036_drug_match_confidence`) and surfaced on the medicines list screen, only when the match was by name alone: "Nura found this by the name on the label. It did not say how strong it is." (en/ms/zh).
- `tests/test_drug_match_confidence.py`: exact brand+strength scores full confidence and identifies; a wrong strength on a real, single-product generic (paracetamol, 999 mg) scores below the floor and is refused; a high-risk drug matched by name alone is refused too; and — the regression that started this — **growing the registry cannot rescue a wrong-strength match, proven by construction**: the same paracetamol-999mg query is checked against both the fixture as it stands and a copy with unrelated products added elsewhere (simulating growth), and the score is identical in both, because it is a function of the label against the one product its name matched, not of how big the registry is.

## The audit #206 asked for: what else treated `NotIdentified` as "not trustworthy"

`grep -rn "NotIdentified"` and `grep -rn "identify("` over `backend/app`, confirmed independently by the clinical-safety review below:

- **`NotIdentified` is raised and consumed in exactly one place**: `_one_product`, called only from `plan()`. Nothing else catches it or infers anything from its absence.
- **The high-risk-needs-photo check is fully downstream of identification** — `match = _one_product(...)` runs before `needs_photo = match.high_risk and ...` in `plan()` — so a match that fails the confidence floor never even reaches that gate; it is refused earlier. This is strictly more protective than before, not less: the store-level hook (`app.safety.high_risk.refuse_dose_without_label_photo`) is still there underneath as an independent second floor either way.
- **But** (this is the clinical-safety review's real finding, not something I'd caught): "downstream of identification" is not the same as "the strength was verified". A high-risk drug's label photo proves a photo exists, not that the strength on it was read — so a bare-name match on warfarin/an insulin/digoxin/methotrexate/an opioid was passing the 0.8 floor at 0.9, uniformly with any other drug. **Fixed**: a high-risk product matched by name alone now scores 0.5, below the floor, and is refused until the label actually states a strength.
- Three other `identify()` callers — `app/safety/health_words.py`, `app/reasoning/visits/summary.py`, `app/delivery/feed/compose.py` — never write a dose or a line; they only ask "does the register know this name" for text detection and feed copy, on generic/brand lookups with no strength given (so `_confidence` can only return 0.9 or better for them). Unaffected, and now documented as exempt by design in the port's own docstring (`app/drugs/registry.py`), per the review's minor finding.

## What I deliberately did not touch

- **The `digoxin` stopgap in #202's tests.** #206 asked me to reconsider it once real scoring exists: with scoring in place, would `ibuprofen 200mg` be refusable "on its own merits" again? No — in #202's grown, 233-product registry, `ibuprofen 200mg` is a genuine exact brand+strength match (Brufen/Advil 200mg, both real registered products), so it legitimately identifies at full confidence; it isn't a near-miss, it's a correct answer. Reverting the digoxin swap would just break that test again. I'm leaving #202's fix as-is and recording this reasoning here rather than touching a separate, already-green PR.
- **A live write-back from the pharmacist queue into registry confidence weights.** Out of scope for this issue; the floor and the scoring formula are the fix #206 asked for.

## Gates

Per standing policy: targeted subset locally, full suite on CI.

- Targeted backend subset — **442 passed**: `test_drugs.py`, `test_drug_match_confidence.py`, `test_medicines.py`, `test_medicines_api.py`, `test_medicine_story.py`, `test_high_risk_conformance.py`, `test_data_map.py`, `test_migration.py`, `test_row_scope.py`, `test_web_seams.py`, `test_family_pushes.py`, `test_feed.py`, `test_feed_formats.py`, `test_red_flag_escalation.py`, `test_visits.py`.
- `make lint` (ruff + mypy) — clean, 262 source files.
- `make plain-words` — 0 failures (94 pre-existing notes, none new).
- `make language` — 0 failures (89 pre-existing notes).
- `make web-test` — 328/328 passed.
- `make build-web` — clean.
- `make checkpoint N=6` on my own port (8211, then re-run on 8212 after the high-risk fix) against a fresh dev server — passed both times.
- Branched off current main (`3ffb911`); confirmed still current before pushing, no merge needed.

## clinical-safety-reviewer report (verbatim, before the two post-review fixes below were made)

> ### Checklist
> 1–5, 7, 8: **PASS** (full reasoning per item omitted here for length; see the review transcript on request — each traced to the specific unchanged code path).
> 6. **High-risk drugs require a label photo before a dose is saved.** PASS mechanically, but see the required change below.
>
> ### Deep-dive findings
>
> **Finding A (required change, fail)** — the confidence formula is drug-class-blind, and the "high-risk is protected downstream" claim is weaker than the diff's docstrings suggest. `_confidence()` returns `NAME_MATCH = 0.9` for *any* exact name match with no strength given — warfarin, insulin, digoxin, methotrexate, an opioid, or paracetamol, identically. I traced the ordering claim through `plan()`: `_one_product(...)` does run before `needs_photo = match.high_risk and artifact.kind is not ArtifactKind.PHOTO`, so the ordering is technically true. But `needs_photo` and the label-photo hook only check that a photo artefact exists — they never check that a strength was extracted from that photo or confirmed against the product. So a label for a high-risk drug that names only the brand (photo attached, strength omitted) clears both gates and a `MedicationLine` is written with `registry_confidence=0.9` without the strength ever being checked. **Required change**: either drop unqualified name-only matches on `high_risk` products below `CONFIDENCE_THRESHOLD`, or require a non-null strength as part of the high-risk label-photo rule.
>
> **Finding B (note, not blocking)** — other `identify()` callers (`health_words.py`, `visits/summary.py`, `feed/compose.py`) are not confidence-filtered, but this looks intentional and low-risk: none write a dose, all ask a boolean/best-match question on generic/brand lookups only. Worth a one-line note in the registry's own docstring so a future reader doesn't mistake this for an oversight.
>
> **Finding C** — no staleness/caching bug: `DrugMatch` is frozen, `_confidence()` is computed fresh on every `identify()` call, and `reconcile()` recomputes `plan()` rather than reusing an earlier `Plan`. PASS.
>
> **Finding D** — `LabelIn.confidence` (the person's yes) vs `DrugMatch.confidence` (the register's own certainty) is well-disambiguated in code and docstrings. PASS.
>
> ### Required changes
> 1. **fail** → **done**: high-risk products now score below the floor on a bare-name match (`HIGH_RISK_NEEDS_STRENGTH = 0.5` in `app/drugs/fixture.py`; new test `test_a_high_risk_drug_by_name_alone_scores_below_the_floor`).
> 2. **note** → **done**: `app/drugs/registry.py`'s docstring now names the three exempt callers and why.

## plain-words-reviewer report (verbatim, before the fix)

> ## Verdict: FAIL (all three languages)
>
> The automated checker missed this — rules 8 and 9 are explicitly the reviewer's job, not automatable.
>
> **1. Rule 13 violation** — this exact screen already has a plain-words term for "strength": `strengthLabel` = "How strong it is" (en) / "Berapa kuat ubat ini" (ms) / "这个药有多强" (zh), used live in the same add-medicine form. The original string used a different word for the same idea in every language — worst in Chinese, where "分量" (amount/portion) is a different word from "强" (strong) in the established term.
>
> **2. "Matched" is backend jargon leaking through**, and a register-shift from the sibling lines on the same card (`sureYes`: "You said yes to this.", `sureRead`: "Nura read this clearly.", `disputed`: "Someone said this is not right." — all plain, personifying, everyday verbs). "Matched"/"padankan"/"配对" are search-algorithm words nobody's daughter says.
>
> **3. Rule 9 failure** — the string only ever fires for the benign case (no strength given at all; a *wrong* strength never reaches the UI, it's refused earlier). As originally written it read like the alarming case, raising a worry ("not the strength") without resolving it or saying whose gap it is.
>
> **Rewrite proposed**: "Nura found this by the name on the label. It did not say how strong it is." (+ ms/zh equivalents reusing `strengthLabel`'s own phrase). — **Adopted verbatim** (en), and adapted with the same reasoning for ms/zh, rendered as two lines rather than one string carrying two sentences.

## Risk

Same standing note as #202: I am not a pharmacist. The scoring thresholds (0.9 for a bare name match, 0.4/0.5 for a wrong strength/form, 0.5 for a high-risk bare-name match) are my own numbers reusing an existing floor, not a clinically-derived scale — reasonable on their face and reviewed, but worth a second look from someone who owns the clinical-safety sign-off before this reaches a real patient.


ashfrnndz21:
**Operator merge note.**

Merging. Independently reviewed: **MERGE**, no blocking items.

This closes #206, and it is the owner's own answer rather than either option I had framed. Asked whether the label check should inspect the label's content or stay a registry miss and be renamed, they said: *"when it recognizes the med, it should give it a confidence score."*

That was the better answer. The defect was never that identification is registry-based — it is that **a match was binary**, so a garbled label colliding with a real product was indistinguishable from a clean, exact one. A score makes those two different things again, and it *improves* as the registry grows rather than degrading: more products means more near-misses, and near-misses score low. The 233-product fixture exposed the weakness precisely because it created the first near-misses.

It also makes this the third use of one rule already in this codebase, which matters more than the fix itself: `CONFIDENCE_THRESHOLD = 0.8` holds back a field scanned off a document photo; since #195 the same floor gates a WhatsApp voice transcript before it may write a Taken tap; and now a drug match. **Nura does not act on a recognition it is not sure of** — one rule in three places, not three unrelated checks.

**The review found the fix closes a worse bug than either PR claimed.** Before this, when a label's stated strength matched nothing on file for that brand or generic, the code fell through and **returned the product anyway, unfiltered by strength**. A paracetamol label misread as "999 mg" would have been silently accepted as the real 500 mg product. That is now refused at 0.4. Not a hypothetical closed by a score — a concrete wrong-strength-accepted bug.

**And building it surfaced something nobody had seen.** The builder's own clinical-safety pass found that a high-risk drug — warfarin, insulin, digoxin, methotrexate, opioids — matched **by name alone** was scoring 0.9, exactly like any other match. The reasoning in the fix is the good part: a label photo proves *a photo exists*, not that the strength was read. So a high-risk product now requires a stated strength to identify at all. That gap was invisible while matching was binary; adding a score is what made it visible. The review confirmed the rule is real, consulted on the identify path, and correctly layered with the pre-existing label-photo gate — the two catch different failure modes and neither makes the other redundant.

**Verified rather than assumed by the review:** the score genuinely discriminates (1.0 for a registration number or exact strength-and-form, 0.9 for a bare name with nothing to check, 0.4 wrong strength, 0.5 wrong form, 0.5 high-risk bare name) — not a constant that always clears the floor, which would be worse than no score at all. The "growing the registry cannot rescue a bad match" claim is structurally true for the wrong-strength and wrong-form class, because `identify()` filters candidates by exact name equality, so unrelated new products never enter a query's candidate set. The `NotIdentified` audit was re-run independently by grep: raised only in `app/medicines/service.py`, and the three other `identify()` callers never check confidence and never write a dose.

**Honest limit, recorded rather than glossed:** the original #206 regression — `ibuprofen 200mg` going from "no product" at 17 items to a correct product at 233 — is *not* what scoring fixes, and the PR says so. That one was a correct answer, not a near-miss. Its permanent test fix still rests on `digoxin`'s deliberate absence from the fixture, which is the same registry-lookup-miss mechanism #206 worried about. Narrower than the framing, and said plainly.

Follow-up #211 is filed and deliberately not fixed here: a person sees the same "Nura could not find this medicine." whether the drug is genuinely unknown or whether a high-risk drug was recognised by name and only its strength was missing — with no hint to retake the photo so the strength is in frame.

#202 follows this one; its migration `0037` chains off this PR's `0036`.

PR #194, opened 2026-09-16
D1 Design pass: the design system, persona navigation, and every screen as designed

ashfrnndz21:
Stage 2 of the design pass, and the owner's reset of the navigation.

## What changed

**The chief's screens speak about Pa, never in his voice.** That was the fix to make first, and
it is now a rule the suite holds: a sweep walks every screen a chief can reach in the caregiver
density — each tab, each place in his Record, each part of Family, the feed, the emergency card,
the symptom log and the pill — and fails on any second-person line about his record. It found more
than the review had: the ten blood-test names ("Your cholesterol"), the blood pressure title,
"Tell Nura how you feel", "What you feel", "Say what you feel…", "You said yes to this.", and — from
the backend, not the catalogue — "I have more at home.", which was his own voice on her screen.
Each now has a twin said about him by name, in all three languages. The pill reads "Pa is not
feeling well" and opens the same flow.

The lines that say "you" and mean *her* stay, and the sweep names each one and why: the feed's own
words for the reader, the box where she writes her own message to the family, and the one line that
asks whoever is holding the phone to describe, in their words, what they are reporting.

**One app, one account, one tab set.** Per the owner's decision and `docs/product-reset.md` §1 and
§6, the two persona tab sets are gone. The tabs are one list — Today, Medicines, Papers, Visits,
Family — the same for the owner of the papers and for anyone holding a key. What the density
changes is the type scale, the target sizes, one thing per screen and how much is shown; it does
not change the tabs. What a key opens still shapes the bar, because a tab whose surface a key does
not open would only reach the backend's no.

**A switcher in the header, on every screen, in both densities.** It names whose papers are open,
so someone acting on his record is never in doubt whose it is. Opening it reads the doors again, so
a key granted a moment ago is there without signing out. For a person who has no papers of their
own — sitting at the doors when the key is cut — the doors learned to look again.

**Family is where access is granted.** The circle carries the one Plum button on the screen, "Give
someone a key", with the role, the parts and the window, in the backend's words for exactly that
person; the rest of Family's parts are rows, one tap from the tab.

**The day's four screens are in the shell** — not feeling well, what to do now, the symptom log and
a tapped feeling — restyled onto the kit's paper tiles and pill buttons, each with the tab bar
under it. They keep no tab bar while they are listening, so a tab tap cannot throw away a symptom
report part-said.
## The map: every screen, and how far it is

**One tab set, the same for everyone.** Density changes the type scale, the target sizes, one
[...]

Issue #208, opened 2026-09-16
The web suite can be red and a PR still merges: 'web' is not a required check


ashfrnndz21:
## What is wrong

Branch protection on `main` requires three checks:

```
["backend", "backend-postgres", "image"]
```

**`web` is not among them.** That job is the one that runs the web unit tests, the build, the plain-words check on the web catalogue, and the entire Playwright suite — every end-to-end walk of the app a person actually touches, and every accessibility check.

So the whole web suite can go red and a PR still merges. Nothing stops it reaching `main`, and nothing tells anyone afterwards.

## Why this surfaced now

Three things today, none of which would have been caught by the required three:

- #199 merged-ready with `web` red on `emergency.spec.ts` — the emergency card's own walk.
- The design pass (#194) found the emergency card was not one tap from her Home, and that **an emergency-only key had no way to reach the emergency card at all** — a key whose only purpose is that card. Web-only; invisible to the required checks.
- The same PR found that at 200% text the demo banner collapsed the scrolling region to **zero** and pushed the tab bar off the bottom of a 360×640 phone. Also web-only.

This is an app whose users are elderly, whose text will be set large, and whose most important screen is an emergency card. The suite that covers exactly those things is the one that cannot block a merge.

It is the same shape as #192 (the checkpoints never run in CI) — a real gate that is not wired as a gate — except worse, because here the job *does* run, reports honestly, and is simply not listened to.

## What to do

Add `web` to the required checks.

Before flipping it, do the work that makes it stick, or it will be reverted the first time it blocks something:

- **Establish the true flake profile.** Today's e2e failures have been a mix of real defects and contention artefacts from several builders sharing one machine — CI is quieter, but prove it rather than assume it. Run the suite on CI several times against an unchanged `main` and record what moves.
- **Fix or name every currently-red test first.** A required check that is red on arrival trains everyone to use the admin override, which is worse than not requiring it. Anything genuinely waiting on other work should be an explicit, named `test.fixme` pointing at its issue — not a silent failure.
- **Keep `strict` false**, as the other three are, so a PR is not forced to re-merge `main` on every unrelated commit.

## Do not

Do not make it required by trimming the suite — dropping the accessibility checks or the slower e2e walks to get a green wall. The value is precisely in the parts that are slow and awkward: the emergency card, the offline behaviour, the large-text layouts, the overlap hit-tests.

PR #213, opened 2026-09-16
Fixes main's red web check

ashfrnndz21:
Fixes main's red web check

## What was asked

`main` (`b469f92`, E22-03 / PR #199) had its `web` CI job red on one e2e test:

```
tests/e2e/emergency.spec.ts — the emergency card, one tap from Today and from Me…
Error: locator.click: Test timeout of 60000ms exceeded.
waiting for getByTestId('emergency-card').getByTestId('print-card')
```

The leading hypothesis going in was that #199 made a card's spoken twin render
**synchronously inside `refresh()`**, before the response returns, and that the added
per-request latency was starving the emergency-card print flow — and, once the same commit's
rerun came back green, whether that latency had made the whole suite's 60s budget marginal
**across the board**, not just in one run.

## Diagnosis — plainly, all the way through

**What is true, confirmed by reading the code:** `_say_ahead` (`backend/app/delivery/feed/compose.py`)
does run synchronously at the end of `refresh()`, awaited before the response returns, for
every new card on every feed refresh (`GET /profiles/{id}/feed`, `.../feed/today`, and the
WhatsApp morning card). This predates #199 — the old code already called `voiced()`
synchronously in the same place; #199 wrapped each item's render in its own
`nested_unit_of_work` savepoint and added an audit-trail write on failure. So "before the
response returns" is not new; #199 added bookkeeping around an existing block, not the
blocking itself.

**Reproduction: could not force the original failure.**
- The exact test, fresh backend + fresh build, `CI=1` (no server reuse, matching CI): passed
  5+ times, both on `b469f92` and on the commit immediately before #199 (3.2–9.8s per run).
- **The failing assertion in the CI log is not the first Print click** — it's the second one,
  spec line 69, inside the offline-reload block. Tracing that path, `refresh()` is never
  reached while offline (`load()` throws on `nura.profile()` before `refreshCard()` runs), so
  `_say_ahead` cannot be in the critical path of *that* click at all.
- The operator's rerun of the same commit came back green — `web` now shows
  `conclusion: success` on `b469f92` via the GitHub API. Main is green right now.

**Cross-checking against #194 and #209, per your ask.** Both have `web=fail` right now.
Neither is the same test:
- **#194** (`web` run [35100383524](https://github.com/ashfrnndz21/NuraApp/actions/runs/35100383524)):
  fails `tests/e2e/a11y.spec.ts:376` — "the writing at 200%… nothing drawn over a line" — a
  static layout/zoom overlap check (`Error: a feed card: drawn over`, 8 elements found covering
  each other). Not a timeout, not a network race, not the emergency card. #194's base **does**
  include #199 (`b469f92` is an ancestor of its base).
- **#209** (`web` run [35099076893](https://github.com/ashfrnndz21/NuraApp/actions/runs/35099076893)):
  fails `tests/e2e/day.spec.ts:177` — a red-flag test — on `expect(received).toMatchObject(expected)`:
  it captured "the next request" expecting `POST .../not-feeling-well` and got
  `GET .../emergency-card.html` instead. This *is* a request-ordering race that happens to
  touch the emergency-card endpoint, which sounds like exactly the systemic symptom you were
  asking about. **But #209's base predates #199** — `git merge-base --is-ancestor b469f92 <209's
  head>` is `no`; its base is `3ffb911`, the commit right before #199. So this exact class of
  flake (the background emergency-card refresh racing a test's own network-capture) already
  existed before #199 shipped. It cannot be #199's doing.

So: three CI failures this week, three different tests, three different failure shapes, and
the one that most resembles a systemic timing issue predates the change under suspicion. That
is evidence *against* one systemic cause, not for it.

**Measured, not inferred: what #199 actually costs a feed refresh.**
I wrote a throwaway pytest script (not part of this PR) that calls `refresh()` directly with
the fixture voice + a real `LocalObjectStore` (exactly what dev/CI run) against a freshly
seeded profile, timed with `time.perf_counter()`, alternated against a "voice off"
baseline (`engine.voice=None`) six times each to average out noise, then ran the identical
script against the commit before #199 for comparison. Numbers, for a refresh that makes 6 new
cards (a representative first-load count):

| | mean OFF | mean ON | delta (6 cards) | per card |
|---|---|---|---|---|
| before #199 (`3ffb911`) | 356.9ms | 369.7ms | +12.8ms | +2.13ms |
| at #199 (`b469f92`) | 186.0ms | 193.5ms | +7.6ms | +1.26ms |

(A second refresh on the same profile — the common case after the first load, since dedupe
means no new cards to voice — costs `_say_ahead` about 19.5ms total, most of it one no-op
savepoint-free early return.)

Both before and after #199, the cost of pre-rendering N new cards' audio with the fixture
voice and a local store is on the order of **one to two milliseconds per card** — noise-level
against run-to-run swings of 30–100ms in the "off" baseline alone, and #199's number is not
even larger than the pre-#199 number in this clean, back-to-back run (an earlier, noisier
run showed a 1.1-second outlier on the baseline side — a GC or disk blip, not signal; excluded
on rerun). For a typical first load this is single-digit milliseconds added to a request that
CI shows taking, in aggregate, ~200–400ms. It is real, and it is not free, exactly as you said
— but it is not a plausible source of a request that's meaningfully closer to a 60-second
timeout. **#199 did not materially slow down `refresh()`.**

## Conclusion

There is nothing here beyond CI-runner variance. Three unrelated flakes surfaced close
together in the same week on a 10–13-minute, single-worker, 125+-test suite; one of the three
(#209's) is proven to predate the commit under suspicion; and direct measurement shows the
mechanism I was asked to weigh (`_say_ahead`'s synchronous cost) is on the order of a
millisecond per card, not enough to move a 60-second budget. I'm not going to dress that up as
more than it is. If you want a genuinely useful next step, it's making the CI runner or the
suite's parallelism less marginal generally (these jobs already run 10–13 minutes serially) —
not touching `_say_ahead`.

Per "whichever way: checkpoint 20 must keep showing `X-Voice-Cache: hit`" — moot here, since
nothing about pre-rendering changed. `_say_ahead` is untouched by this PR.

**What I did fix**, found while tracing the exact code path the failing assertion exercises
(independent of the flake question above): `saveCard()` in `web/src/offline/emergencyCache.ts`
unconditionally overwrote the phone's kept printable HTML (`KeptCard.html`) with whatever the
latest read produced — including `null`, when only the printable-page fetch
(`GET /profiles/{id}/emergency-card.html`) failed. `readCard()` degrades that one fetch to
`null` by design (a `try/catch`, so the card's JSON lines still get kept even if the page
can't be read) — but `saveCard()` then wrote that `null` over a printable page the phone
already had. A single transient failure of the printable-page endpoint, on any background
refresh (the Today screen's once-a-day re-read, or reopening the card), would silently take
away the Print button offline, even with nothing actually wrong with the card. I'm not
claiming this explains the CI failure (the trace shows `refresh()`/`refreshCard()` isn't even
reached in the failing click's code path) — it's a genuine, separate bug in exactly the area
the test exercises, worth fixing regardless: the offline copy is the whole point of this
cache, and it should never regress data it already had because of one bad request.

The fix: `saveCard` now takes the previously-kept entry and falls back to its `html` when the
new read's `html` is `null`, **gated on the previous entry being in the same language** — a
gap the clinical-safety review below caught in the first version of this fix (a carried-over
page in the wrong language, paired with a fresh card and a single `fetchedAt` implying both
are equally current). A real `html` from the new read, even an empty string, always still
wins. Both call sites (`Today.tsx`'s `refreshCard`, `Emergency.tsx`'s `EmergencyScreen` effect)
already had the previous entry in scope. Three unit tests cover the fallback, the
still-replaces-on-success case, and the language gate.

## Acceptance, answered line by line

There's no single backlog acceptance line — it's the operator's own brief, in two rounds:
- *Reproduce it locally first* — done; could not reproduce, 5+ clean runs both sides of #199.
- *Find the actual cause, don't stop at the symptom* — traced `kept.html`'s full write path;
  found and fixed a real, independent overwrite-to-null (then wrong-language) bug.
- *If it's pre-render latency, move it off the request path* — did **not**, and now have a
  measured reason not to: ~1–2ms/card is not a latency problem to solve.
- *Check #194 and #209* — done; different tests, different failure shapes; #209's predates #199.
- *Measure, don't infer* — done; table above, methodology described, before/after #199 both run.
- *checkpoint 20 must keep showing `X-Voice-Cache: hit`* — untouched, not at risk.
- *Care: don't fix by extending the timeout, retrying the test, or weakening assertions* — the
  test file is untouched, in either round of this PR.

- No change to `_say_ahead` / the voice pre-render path. Measured, not just argued: not worth it.
- No change to any e2e spec, its timeout, or its assertions.
- No backend changes at all — this PR is web/ only.

## Gates

- `make web-test`: 332 passed (39 files) — up from baseline's 329 (three new tests added).
- `make build-web`: typecheck + `vite build` clean.
- Affected spec (`web/tests/e2e/emergency.spec.ts`, all 4 tests, fresh server, `CI=1`): 4 passed.
- `make lint` (backend, ruff+mypy): not run — no backend files changed (the timing measurement
  was a throwaway pytest script, deleted, never committed).
- `make plain-words` / `make language`: not run — no strings in `web/src/strings/**` changed
  (confirmed by the plain-words reviewer below).
- `make checkpoint`: not applicable — no backend behaviour changed.

## clinical-safety-reviewer

Scope confirmed: offline cache logic for an already-verified, already-rendered emergency card
(`saveCard`/`readCard` in `web/src/offline/emergencyCache.ts`, plus the two call sites and
their unit test). No new data source, network call, drug/medicine logic, consent, or key logic.

1. Profile reads go through the keys module with a key context — N/A, no new reads.
2. Every new Fact carries provenance; every Card records its State — **Fail (now fixed)**: the
   fallback wasn't gated on language, so a carried-over printable page could show the wrong
   language's words under a fresh-language card, both under one `fetchedAt` implying equal
   freshness. *Required change, applied*: only carry `html` forward when
   `previous.card.language === read.card.language`; otherwise treat a stale-language `previous`
   as none. Verified fixed in the second commit.
3–6. No medicine, drug, red-flag or high-risk-label logic touched — N/A.
7. Health data stays in region; none reaches analytics — Pass. No analytics call added; data
   still only flows into the existing IndexedDB-backed `kv` store, scoped exactly as before.
8. Consent/binding scoping — Pass (with caveat now closed). `previous` is only ever sourced via
   `loadCard(id, binding)`, which deletes and returns `null` for any entry whose binding
   doesn't match the current key/scope set, so the fallback can't leak a page across keys or
   narrower scopes. Language was the one gap; it's closed.

Follow-up suggestion from the reviewer (not blocking, not applied): consider footnoting the
`fetchedAt` when `html` is a carryover, since ADR 0010 didn't anticipate a fresh-JSON /
carried-HTML staleness mode. Noting it here for the operator's judgement rather than
speculatively adding UI for a case that hasn't happened yet.

## plain-words-reviewer

No patient-facing strings were introduced or changed in this diff. `git diff main --
web/src/strings/en.ts web/src/strings/ms.ts web/src/strings/zh.ts` is empty. The changed lines
are `saveCard`'s fallback logic and its JSDoc, two call-site argument additions, and three
Vitest test descriptions — all developer-facing, none of it something a patient reads or
hears. Nothing to check against `docs/plain-words.md`.


ashfrnndz21:
**Operator merge note.**

Merging. This PR exists because I sent a builder to find a regression that turned out not to be one — and it came back with a real bug instead, and with my hypothesis disproved.

**My hypothesis was wrong, and it was measured wrong rather than argued wrong.** I believed #199 had made feed refresh slower by rendering audio synchronously inside `refresh()`, and that this pushed a 60-second e2e over its limit. The builder timed it: `_say_ahead` costs about **1–2ms per newly-voiced card**, noise-level, and no larger after #199 than before. It then found the detail that settles it — the assertion that actually failed in the CI log was the **second** Print click, the one after an offline reload, and that path never reaches `refresh()` at all while offline. It also checked my claim that #194 and #209 were failing the same test: neither is, and #209's failing race predates #199 entirely.

So it declined to move pre-rendering off the request path, declined to touch the timeout or the assertions, and wrote the flake up plainly as unexplained shared-runner variance on a 125-test single-worker job. That is the right outcome: a re-run of main came back green on the same commit, and a fabricated fix would have left a real cause unfound while looking resolved.

**The actual bug it found while it was in there:** `saveCard()` in `web/src/offline/emergencyCache.ts` was **unconditionally overwriting the phone's kept printable HTML with `null`** whenever a background re-read's printable-page fetch failed — even when the card's own lines came back fine. One transient network hiccup and the offline Print button silently disappeared from someone's emergency card, with nothing to tell them it had gone.

That matters more than its size suggests. The Print button is how an emergency card gets onto paper for a person who cannot use a phone, and the failure mode was silent, transient and invisible — the worst combination, because nobody would ever report it. `saveCard` now carries the previous printable HTML forward when the new read is null.

**Its own clinical-safety pass then caught a gap in that fix**, which is why the carry-forward is gated on matching language: a page carried over could otherwise have shown the **wrong language** under a fresh card. An emergency card in a language the reader cannot read is worse than no printed card at all.

Three unit tests cover it; `web-test` 332 passed, `build-web` clean, `emergency.spec.ts` 4/4.

The flakiness itself remains open and honestly unexplained. The related finding — that `web` is not a required check, so a red web suite merges — is #208.

PR #186, opened 2026-09-16
An urgent alert is never downgraded when its template is not approved

ashfrnndz21:
Closes #174

## What changed

Outside a family member's 24-hour WhatsApp window, only Meta-approved templates may be sent. The red-flag tier notices (ambulance, hospital-now, night — added in #147) are still pending approval, so the code fell back to the approved `red_flag_notice` template ("Call Dr Tan today."), which tells the family a **lower** urgency than the flag's real tier.

`flag_message()` in `backend/app/delivery/triggers/ladder.py` now treats `red_flag_notice_self` and the base `red_flag_notice` ("call {doctor} today") as WhatsApp candidates only when **no tier applies**. Once a tier applies (`tiered is not None`):

- The tier's own template goes if Meta has approved it (unchanged).
- Inside the family member's 24-hour window, the tier's own words still go as free text if the template isn't approved (unchanged — nothing changes inside the window).
- Outside the window, if the tier's own template isn't approved, the **sole** WhatsApp fallback is a new neutral template, `red_flag_notice_urgent` — "This one we do not wait for. {name} is not feeling well. Open Nura now." It states no action and names no doctor, hospital or ambulance.
- Where even that isn't approved, **no WhatsApp goes at all**, and the reason is recorded in the `Delivery` row's `passed_over` (via the existing `_by_whatsapp`/`_every_way` machinery in `deliver.py`, unmodified).
- The app push and the always-written in-app family-page notice (`DeliveryChannel.IN_APP`) are unaffected — they never depended on which WhatsApp candidate succeeded.

Also updates `docs/adr/0010-red-flag-tiers.md` and `docs/trust/clinical-sign-off.md` to match (the "approved notice as last resort" tradeoff they documented is exactly what this fixes), and fixes a couple of pre-existing merge-artifact duplicates in `test_whatsapp_templates.py` (a dead duplicate test function and a duplicate `E19_SIX` assignment) found while updating the template count there.

## Tests

`backend/tests/test_red_flag_escalation.py`:
- `test_outside_her_window_an_unapproved_tier_never_falls_to_call_the_doctor_today` — chest-pain-equivalent (hospital-now) flag outside the window, neither the tier template nor the neutral one approved: no WhatsApp goes to her, the reason is on the trail, and the in-app notice is still written.
- `test_outside_her_window_with_the_neutral_notice_approved_it_goes_not_the_tier` — the neutral template goes when approved, never the tier's own words or "Call Dr Tan today."
[...]

PR #167, opened 2026-09-15
B3 A whole day on WhatsApp by the scheduler; consult recording in resumable chunks

ashfrnndz21:
Closes #41, closes #129, closes #155

## What this does

**#41 (E11-01): a whole day on WhatsApp, run by the scheduler.**
- The daily check-in to Pa and the chief's evening family notice are two E11 trigger rules (`backend/app/delivery/triggers/day.py`, registered in `rules.py`, run from `engine.run_due`). `run_family_notice` is gone from `level0.py`.
- The check-in goes at his check-in time from his settings (otherwise 10:00), once a day, never late. The notice goes to each chief at 20:00: a count of what was written down this week that her own key opens, never content.
- Both run under #121's one daily cap per person, the quiet hours and the recipient's channels. Neither goes on a day with an open red flag (the hold is logged once). The check-in is not asked twice: not after he said how he is, not after the day's nudge asked, not while a yes/no question of his is open.
- Both show in #137's delivery log, named, with the reason when held, in en/ms/zh. A chief sees only her own notice; the "red flag" hold shows only to a key that opens the emergency lines.
- `backend/tests/test_whatsapp_day.py` drives the whole day through `POST /dev/clock` and `POST /dev/run-triggers` on the frozen clock; everything Pa sends comes in through the signed webhook.

**#129: the consult recording uploads in resumable chunks.**
- Server: `backend/app/ingestion/chunks.py`, `backend/app/channels/api/recording_uploads.py`, ADR 0013. One new table, `consult_upload`.
- Each chunk goes through #135's `read_capped` (1 MiB, 413 `ChunkTooLarge`), asks the RECORDING gate again, and lands in the region's store (`guard_region`). The server puts the chunks together in order and hands them to #128's `record_consult`: the same bytes, speakers and clips as a single upload.
- **Nothing is stored before the doctor's yes.** The server refuses a chunk until his yes is on the upload (`NoYesFromTheDoctor`, 409); the phone keeps the audio in its memory until then. A no, or no answer, leaves nothing in the store. After the yes, leaving the page throws away every chunk sent; the engine's five-minute run throws away what the phone could not (no answer in 15 minutes, not finished in 2.5 hours, no RECORDING consent).
- Hearing stays family-only (`OnlyTheFamilyHears`, unchanged).
[...]

PR #215, opened 2026-09-16
An answer is an answer: any template that asks how he feels opens the window to answer it

### LINKED ISSUE #205: He answers the check-in on WhatsApp and is told there is no question open
## What happens to him

Nura asks Pa "How are you feeling today?" on WhatsApp. He replies **"ok"**. Nura answers:

> I have no question open for you.

and nothing of his feeling is written down.

This is on `main` today. It is documented, with the assertion `felt == []`, in `backend/tests/test_whatsapp_day.py`'s scheduler-driven whole-day walk (#167) — the builder found it, wrote it into the test honestly rather than working around it, and flagged it rather than fixing it inside a merge. This issue is that fix.

## Why

`_check_in_open` (`backend/app/channels/whatsapp/inbound.py:1774`) decides whether his "OK" is an answer, and it keys on the **template name**, exactly:

```python
WhatsAppMessage.template_name == "feeling_check_in",
```

Two different things ask him the same question:

- the plain check-in, on the `feeling_check_in` template;
- the day's smart nudge (W7, E17-03), on the `nudge` template — and its last line is literally `"How are you feeling today?"`.

When the nudge is what asked — which is what happens on a day where a change of his is worth asking about, so the more interesting day — his answer falls through to the general reply. The engine deliberately stands the plain check-in down so he is not asked twice; the consequence is that on exactly those days his answer is thrown away.

## Why it matters more than it looks

He did the thing we asked. He answered, in one word, on the channel we chose for him, and Nura told him there was no question open. For someone whose whole relationship with this app is a few words on WhatsApp, being told his answer was not a thing he was asked is worse than not being asked at all.

And the data loss is silent: no refusal, no flag, no trail line saying a feeling was offered and dropped. It looks from the record exactly like a day he did not answer — which is a signal the nudges and trends both read.

## What to build

Make "is a question of his open?" a fact about **what was asked**, not about which template carried it. Some options, and the choice matters:

- Mark the outbound message itself as one that asks for a feeling — a column or a `why` key on the delivery row — so any template that asks the question opens the answer window, and a template added later inherits the behaviour instead of silently missing it. This is the shape #186 taught us to prefer: a prefix test rather than a hard-coded list, so a new member of the family cannot be born outside the rule.
- Or make the nudge of kind `check_in` write the same marker the plain check-in writes.

Whichever way: the rule must not be a list of template names that a future template can fall outside of. That is the same defect as the red-flag exemption list in #186, and it bit us there for the same reason.

Cover in tests: the nudge asks and he answers "ok" — a feeling is written down; both askers on one day cannot double-count his answer; and his answer to yesterday's question does not land on today's.

ashfrnndz21:
Closes #205.

Nura asks Pa **"How are you feeling today?"** on WhatsApp. He replies **"ok"**. Nura answers:

> I have no question open for you.

and nothing of his feeling is written down.

Found by #167's scheduler-driven whole-day test, which documented it honestly with `felt == []` rather than working around it. That assertion now asserts the opposite.

## Why it happened

`_check_in_open` decided whether his "OK" was an answer by keying on the **template name**, exactly:

```python
WhatsAppMessage.template_name == "feeling_check_in",
```

Two different things ask him that same question: the plain check-in on the `feeling_check_in` template, and the day's smart nudge (W7, E17-03) on the `nudge` template — whose last line is literally "How are you feeling today?". When the nudge is what asked — which happens on exactly the days something of his was worth asking about, because the engine deliberately stands the plain check-in down so he is not asked twice — his answer fell through to the general reply and was thrown away.

## Why it mattered more than its size

He did the thing we asked. He answered, in one word, on the channel we chose for him, and Nura told him there was no question open. For someone whose whole relationship with this app is a few words on WhatsApp, being told his answer was not a thing he was asked is worse than never being asked.

And the loss was **silent** — no refusal, no flag, no trail line saying a feeling was offered and dropped. In the record it looked exactly like a day he did not answer, which is a signal the nudges and the trends both read. The day he answered and the day he ignored us were indistinguishable, so the system learned the wrong thing about him.

## The fix, and the shape of it

"Is a question of his open?" is now a fact about **what was asked**, not about which template carried it: the outbound message itself is marked as one that asks for a feeling (`asks_feeling`), so **any** template that asks the question opens the answer window, and a template added later inherits the behaviour instead of silently missing it.

**What it must not be is a list of template names a future template can fall outside of.** That is precisely the defect #186 fixed in the red-flag exemption — a hard-coded list of three notice names that every tiered notice added afterwards silently fell outside, so the most urgent alerts were being refused with `ConsentWithheld`. Same shape, same trap, one layer over. Built so a new member of the family cannot be born outside the rule.

## A classification called out deliberately

`whatsapp_message.asks_feeling` is classified **OPERATIONAL**, consistent with the precedent at the same table — `template_name` is operational while `catalogue_key` is health. The clinical-safety reviewer passed it and raised a nuance worth recording rather than leaving implicit: previously `template_name="nudge"` told an analytics consumer nothing about content; now `asks_feeling=True` on a nudge row says "this nudge asked the feeling question." It still discloses nothing about his actual feeling, and the prompt is identical for everyone — so it passes on the rule, but it is a **narrow widening** of what an operational row says about message content. Flagged here as a conscious call, so anyone who wants it stricter finds the reasoning rather than a silent default.

## Gates

Merged current main (`563dd81`). Migration `0039_feeling_question_marker` chains off `0038_dose_taken_late`; `alembic heads` prints exactly one head. The data-map block was regenerated from `render()` after the merge, not hand-resolved, and `python3 -m scripts.data_map --check` is clean.

- `pytest tests/test_whatsapp.py tests/test_whatsapp_day.py tests/test_nudges.py tests/test_triggers.py` — **70 passed**
- `pytest tests/test_data_map.py tests/test_migration.py` — **8 passed**
- `ruff` clean; `mypy` no issues in 264 source files
- `make plain-words` — 8010 strings in 33 files, **0 failures**
- `make language` — 8149 strings, 2745 keys in 31 catalogues, **0 failures**

The full suite is deferred to CI by standing policy. Both reviewer agents were run over the diff; the clinical-safety pass verified the key-context reads and writes are unchanged and still audited, the Fact provenance path is untouched, the red-flag check still runs before `_check_in_open` in `_answer`'s control flow, and consent gating in `_may_message` is not bypassed. The plain-words pass found one test fixture reading "You told Nura about tired last week." — a variable name dropped into a sentence template — now "You told Nura you felt tired last week."


ashfrnndz21:
**Operator merge note.**

Merging. Nura asked Pa "How are you feeling today?" on WhatsApp, he answered "ok", and Nura told him *"I have no question open for you."* — his feeling never recorded, silently. `_check_in_open` keyed on `template_name == "feeling_check_in"`, but the smart nudge asks the identical question on the `nudge` template, so on exactly the days something of his was worth asking about his answer was thrown away. In the record that looked the same as a day he ignored us, which the nudges and trends both read.

Now "is a question of his open?" is a fact about **what was asked**, not which template carried it: the outbound message is marked `asks_feeling`, so any template asking the question opens the answer window and a future template inherits it. Deliberately not a list of template names — that is the shape #186 fixed in the red-flag exemption, where every tiered notice added after the original three silently fell outside the list and was refused.

`asks_feeling` is classified operational, consistent with `template_name`; recorded in the PR as a conscious call, since a nudge row now reveals that it asked the feeling question (nothing about the answer).

Migration `0039` chains off `0038`; the data map was regenerated from `render()`, not hand-merged. All four CI checks green.

Issue #214, opened 2026-09-16
The Family consents screen quotes his consent wording verbatim, unattributed — she reads "your" as herself


ashfrnndz21:
## What happens

`ConsentsPart` (`web/src/screens/family/Consents.tsx`) titles the screen caregiver-aware —
`whose(here, words.consentsSelf, words.consentsOther)` — but the wording of each consent is not.
`wordingLines(consent)` renders `consent.wording` verbatim: the literal sentences Pa read and
agreed to, sourced from `backend/app/consent/texts.py`'s `SHARE_WITH_PERSON` template.

On Mei's own phone, under Family, this is what she sees:

```
You are letting Mei, your daughter, see some of your record.
- your medicines
- your visits to the doctor
- your blood pressure book and your sugar numbers
- your papers
- your private notes
- your insurance letters
- your emergency card
- your family list
- your questions to Nura
Mei can see them until you say stop.
```

Nothing on the screen marks this as a quotation, or says it is a copy of the words *he* read
rather than a description of what is happening to *her*. She reads "your medicines" the only way
a second-person sentence addressed to her can be read: as hers.

## Why it matters more here than an ordinary copy slip

This is not decorative copy — it is the **consent record**, the one screen whose entire job is
being unambiguous about who agreed to what, for whom, and under what scope. A caregiver misreading
"your medicines" as her own medicines, on the exact screen that exists to answer "what did he
agree to, and by whom," defeats the screen's own purpose. It is the worst place in the app for
this class of confusion to happen.

## How it was found

The broad sweep added in #194 (`no caregiver-density screen says a second-person line about his
record`, `web/tests/e2e/about-him.spec.ts`) failed on it. This screen has presumably been
ambiguous to every caregiver reading it since it was built — the sweep only reached a correct,
settled read of it once #194 repaired the sweep's own settle-timing (its PR body: "a screen can be
read before it has said anything and pass by saying nothing"). It is carved out of the sweep here,
narrowly and named for this issue — the same way #210's two backend paths are carved out — not
exempted as a category. A verbatim quotation is not inherently exempt from the rule; *this*
quotation, unattributed, on someone else's screen, is a real defect.

## The fix (either is enough; both is better)

1. **Attribute it visibly.** Render the wording under a line that names whose words they are —
   the same pattern the printed record already uses ("These are the words {giver} read…",
   `backend/app/consent/export.py`) — so "your" is read in a quoted voice, not Mei's.
[...]

PR #212, opened 2026-09-16
A late 'Taken' is recorded as late, its pattern is visible, and whoever the ladder reached is told it stood down

ashfrnndz21:
Closes #198

## The owner's decision (quoted from the issue's decision comment)

> **Record it, and record when the reply came.** A Taken that arrives after the window
> closed marks the dose taken — but the reply's own time is stored with it, so a late Taken
> is visibly late in his record and in his trends, rather than indistinguishable from an
> on-time one.

Answered line by line, against the "Build" list in that comment:

1. **"The tap carries the reply's time as well as the dose's anchor, so 'taken' and 'taken
   late' are distinguishable in the row, not inferred by comparing two fields at read time."**
   Done. `DoseTaken.late: bool` (migration `0036_dose_taken_late`), computed once in
   `record_dose_taken` from `app.medicines.windows.is_late` and stored on the row. The
   WhatsApp path now passes the provider's own message timestamp as `taken_at`, not the
   moment the backend happened to process the reply.

2. **"His trends and the pattern signals read it as late. A week of doses taken two hours
   after their window is a real finding about his day and today it looks like a perfect
   week."** Done. The Today card's `Slot.taken_late` (and `SlotOut`/`TakenOut` over HTTP)
   carries the bit through without re-deriving it. The weekly pattern signal reads it too:
   `_late_pattern` (new, beside the existing `_pattern`/`DOSES_UNTAPPED`) tells the one on
   duty when three or more taps in seven days were late — `TriggerType.DOSES_LATE`, template
   `doses_late_count`, rule `three_late_doses_in_seven_days`, same cap and quiet hours as
   `DOSES_UNTAPPED` beside it. Without this, a late-but-taken week never showed in the
   untapped count (correctly — it was taken) and so looked perfect; now it doesn't.

3. **"The card and the trail say it in his own words — plainly that it was written down
   late, never scolding."** Done, in English, Malay and Chinese, `make plain-words` and
   `make language` clean: the WhatsApp confirmation adds one line, once, after a late tap
   ("This was written down later than usual." / "Ini ditulis lewat daripada biasa." /
   "这次记得比较晚。"), and the new `doses_late_count` template says the count plainly ("Pa
   said Taken 3 times late this week.") the same shape as the existing `doses_count`. The
   trail is the `DoseTaken.late` bit itself plus the ladder row, never rewritten.

4. **"Whoever the ladder actually reached must be told it stood down."** Done. Checked what
[...]

PR #177, opened 2026-09-16
F1 The feed, full: clips with captions, local and seasonal alerts, food cards, engagement, the chief's Watching and Sent panels

ashfrnndz21:
The feed's richer formats, end to end (E09-06, E09-07, E11-08, E11-09, E21), against
`docs/health-feed-spec.md` and `docs/ui-mockup-v2.html`.

## What he gets

**Clip cards.** A 20–30 second excerpt of an allowlisted video, narrated in his language with
WebVTT captions, and a still served from Nura's own object store — no third-party embed, no
request to a video site until he taps "Watch the whole video at {publisher}" himself. Where a
licence forbids reuse, he gets the still plus the narration and the captions, and the refusal
that says so is on his trail. Rendering sits behind `ClipRenderer` with a fixture adapter, so
nothing here needs ffmpeg. Nothing ever plays by itself.

**His week in 30 seconds** (E11-09): this week's story cards read back to him. It repeats his
own record, so it carries no boundary line and links nowhere.

**Local and seasonal alerts** (E09-07): dengue, haze, heat, the fasting month, festive food
and travel, from allowlisted health authorities — each says what to do today and ends on its
boundary line. His area is coarse by construction: a district from the region's list, or two
or three digits of a postcode, refused otherwise. It is set on his own yes, classed as an
identifier, and never sent to a searcher.

**Food and habit cards**: one concrete choice a week, by condition and season.

**Engagement** (E11-08): a queue on the phone (opened, played with seconds, replayed,
dismissed, asked-more, shared) flushed to `POST …/feed/events`. After two unopened text cards
the feed goes voice-first. Time in the feed is not measured and not learned from — there is a
test that greps the whole tree to prove `seconds` is written and read by nothing that orders
the feed.

**The chief's panels**: "Watching for Pa" (each watch, its sources and cadence, which she can
add or pause) and "Sent to Pa this week" (each card with status, source and actions), over
`GET /search-jobs` and the add and pause routes. **Search chips**: Records, Web, Providers,
Videos — Web and Videos from the allowlist only, and the query travels in a POST body so his
words stay out of the URL.

**The caregiver feed has no gate.** The duty card was being treated as one, so it lost its
side actions; it is one of her today cards now and carries every action, "Not for me" and
"Hear" among them.

Richer demo fixtures in en/ms/zh, all sourced: explainers for blood pressure, diabetes,
cholesterol, warfarin, amlodipine and metformin; two clip cards; a Penang dengue alert; a haze
[...]

PR #216, opened 2026-09-16
fix: unbreak main's web job — red-word test ordering + a caregiver twin for the metrics table

ashfrnndz21:
## What this fixes

`main` (`563dd81`) was red on its `web` job with two failures, both appearing only after #212 and #194 merged in sequence (each green on its own CI run).

## Diagnosis of the red-word failure — `day.spec.ts:177` — stated plainly

**The #191 guarantee is intact. It is not broken.** The test's assertion was wrong, not the product.

Proven two ways, per the ask not to guess:

1. **A deterministic unit test** (`web/tests/unit/client.test.ts`, `describe("the urgent queue (#191)…")`) mocks `fetch` so a background read never resolves except by abort. If `api/client.ts`'s `enqueue()` ever failed to abort a non-urgent, already-on-the-wire read for an urgent call, this test hangs and times out — no wall clock, no browser, no CI flake surface. It passes. A manual mutation (commenting out the abort line) makes it hang exactly as expected, confirming the test isn't vacuous.
2. **Manual e2e instrumentation**: holding `emergency-card.html` open indefinitely via `page.route` while the red word is said shows the urgent POST completes in ~130ms regardless — the held GET is aborted, never waited on.

**The actual bug was in the test's assertion**, and it went through two drafts before landing correctly (see commit history on this branch — the operator caught the first draft weakening the wrong thing):

- **Draft 1 (wrong)**: `expect.poll(trail.inflight).toBe(0)` samples a momentary gap between Today's own serial background reads (state, medicines, feed, …, `emergency-card.html` — `useToday`'s `refreshCard`, the area #213 and #194 touched) before capturing `mark`. In that gap, the next background read can *start* before the click's async work (`recorder.stop()`) finishes and the urgent POST is dispatched — so "the very first `start` after mark" is occasionally a GET, even though it's aborted almost immediately and the POST is never delayed. My first fix replaced the ordering check with `some()` (did the POST happen at all) plus a 3-second wall-clock bound — which is exactly the "loosened into meaninglessness" failure mode: it would pass even if the red word genuinely queued behind every background read on the page.
- **Draft 2 (correct, what's in this PR)**: kept the relaxation of "must be the very first *start*" (that part of the old assertion really was racy and not what #191 promises), but restored the actual guarantee exactly: **no other request is allowed to run to a real, successful completion before the red word's own completion.** `apiTrail()` now tags every `"end"` with `ok` — `true` for `requestfinished` (the exchange ran to a real response), `false` for `requestfailed` (aborted — including the urgent queue's own abort, which is the mechanism working, not a violation of it). Both tests locate the red word's own POST by method+path, find its own `end`, and assert nothing before that `end` is another request's *successful* end. No `Date.now()` bound.

Same fix applied to the sibling `feeling cloud` test (`day.spec.ts`), which had the identical pattern and the identical latent race — not currently reported by CI, but the same mechanism, so left unfixed it would have flaked next.

## Failure 2 — the caregiver-voice sweep, `about-him.spec.ts:98`

Family Metrics screen (`web/src/screens/family/Metrics.tsx`) rendered `kinds.presence` ("Thinking of you") as a table row label, on Mei's screen — the exact same string that correctly reads as her own message when it stands alone in the composer (`Messages.tsx`'s `templates.thinking_of_you`) reads ambiguously in a table of counts about him ("Thinking of you\t0\t0\t0").

Same class of finding as #214. Fixed properly, not carved out, by following the codebase's existing convention for this: the `*_THEIRS` twins in `backend/app/delivery/strings.py` (read by `app/channels/about_him.py`) — a caregiver-safe, third-person, name-anchored rewrite alongside the original. Added `kindsTheirs` to `web/src/strings/{en,ms,zh}.ts` (all six kinds: anticipation, check_in, pattern, commitment, recognition, presence) and `types.ts`, and pointed `Metrics.tsx`'s table at it via `fill(..., { name: here.papers.display_name })`.

Checked the sibling `templates` block per the instruction: it's used only as standalone pill labels in the message composer, already correctly excluded by the sweep's `HERS` regex (`/^Thinking of you$/`, matched as a whole line). No second instance of the problem there.

## Gates (all clean, fresh `dev.db`)

- `make web-test`: 373/373 (374 with the new unit test counted separately — see below)
- `make build-web`: clean
- `make lint`: ruff + mypy clean (264 source files)
- `make plain-words`: 8028 strings, **0 failures**, 120 pre-existing notes (none touching new lines)
- `make language`: 8167 strings, **0 failures**, 89 pre-existing notes (none touching new lines)
- `about-him.spec.ts`, `day.spec.ts`, `emergency.spec.ts`, `feed.spec.ts` in full: **31/31** (3 pre-existing `test.fixme` skipped, unrelated to this PR — #177/#214 carve-outs already on `main`)
- The two originally-failing tests, repeated **50/50** with no flake on a freshly reset database (an earlier 409 `ProfileAlreadyOwned` in one repeat run was this session's own accumulated manual-testing data colliding on a 4-digit phone suffix — not a real issue, and not something CI's ephemeral database would ever hit; resolved with `make reset-db`)

No migration, no new column — `test_data_map.py` not applicable.

## What I deliberately did not do

- Did not touch `templates` (checked, not broken).
- Did not apply the same ordering-assertion hardening to every other e2e test that touches the urgent queue elsewhere in the suite (e.g. symptom log) — out of scope for this fix; flagged as a candidate follow-up if useful.
- Did not investigate or fix the pre-existing flake in `nudges.spec.ts:171` ("Response has been disposed" — a `route.fetch()`/disposal race, unrelated to anything in this diff, reproduced independently of this branch, not in the failing-spec list this PR was scoped to).

## Risk

Low. The product code (`api/client.ts`) is untouched — this PR is test-only plus two string catalogues and one screen component reading a new string key. The new unit test is a genuine regression guard for the abort mechanism (verified it fails/hangs on a mutation). The two e2e assertions are now exact rather than weaker, so this should not introduce new flake — if anything it removes the two known flake sources (the `inflight===0` race and the wall-clock bound I'd added and then reverted).

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


ashfrnndz21:
**Operator merge note.**

Merging. This unbreaks `main`, and the important half of it is a diagnosis rather than a fix.

**The #191 red-word guarantee was never broken.** `main` was failing `day.spec.ts`'s assertion that the first request to start after a red word is the urgent POST — an `emergency-card.html` GET was starting first. That could have meant the most safety-critical behaviour in the product had regressed, so I asked for evidence rather than a judgement, and refused to accept the test being loosened to get green.

It was proven intact two ways. A new deterministic unit test mocks `fetch` so a background read only ever resolves via abort — and the builder **mutated `client.ts` to disable the abort and confirmed the test then hangs**, which is what makes it a real test rather than one that passes incidentally. Then it held `emergency-card.html` open indefinitely while saying the red word: the urgent POST completed in ~130ms regardless.

**What was actually wrong was the assertion**, and the correction matters. A background read may legitimately *start* in the instant before he speaks — that is not a violation, it is the queue's own mechanism about to abort it. But the original test made a second claim that was never racy and is the real guarantee, and the first attempt at this fix dropped it, replacing it with an existence check and a three-second wall-clock bound. That would have passed even if a red word genuinely queued behind every background read on the page, while looking like reasonable de-flaking. I sent it back.

The assertion now says exactly what #191 promises:

> **No other request is allowed to run to a real, successful completion before the red word's own completion.**

That needed an `ok` flag on the test's request trail, so it can tell "the queue aborted a background read" — correct, the mechanism working — from "a background read finished ahead of him", which is the actual violation. Without that distinction the test could not express the guarantee at all, which is presumably why the original settled for the cruder version. It is now **stronger** than what it replaced: the old form would have flagged an abort as a failure.

**The second failure was a real finding, not a flake.** The repaired caregiver-voice sweep caught `kinds.presence` — "Thinking of you" — rendered as a row label on the family Metrics screen. On her screen that reads as if Nura is thinking of *her*, when it names a kind of message sent to *him*. Same class as #214. Fixed properly with caregiver twins following the codebase's existing `*_THEIRS` convention rather than a second mechanism, and the sibling `templates` block was checked for the same problem and is clean. Not carved out — this one is a string fix, not a design decision.

Neither failure was caught before merge because #212 and #194 were each green alone and the **combination** was never tested. That is exactly the hole #208 describes: `web` is not a required check, so a red web suite reaches `main`.

Product code untouched — test-only plus two string catalogues and one screen reading a new key. The two originally-failing tests passed 50/50 repeats with no flake.

PR #218, opened 2026-09-17
CI runner for the checkpoints (#192): a thin `make checkpoints-ci`, plus what it found broken on main

ashfrnndz21:
Addresses #192. **I could not edit `.github/`** (out of scope for this session by the operator's own rule) — this PR is the runner side only: the script, the make target, and proof it works. The workflow YAML that wraps it is at the bottom of this description for the operator to add.

## Acceptance, quoted from #192, answered line by line

> Boot the app the way `make dev` does, on a throwaway database, with the frozen clock (`NURA_FROZEN_CLOCK`) and the fixture providers the dev target already sets.

Done. `scripts/run_checkpoints.py` (wired to `make checkpoints-ci`) shells out to `make dev` itself — same recipe, same fixture env vars (paper/visit/voice/speaker/feed fixtures, the dev code sender, the WhatsApp fixture provider, the pharmacist token) — with three overrides layered on top, the same way `web/playwright.config.ts` already does for `make web-e2e`: `NURA_DATABASE_URL` points at a fresh sqlite file in a `tempfile.TemporaryDirectory()`, `UVICORN_PORT` is a free port the OS hands out (`socket.bind(('',0))`), and `NURA_FROZEN_CLOCK` defaults to `2026-09-14T10:00:00+08:00` (the same instant `playwright.config.ts` freezes at) unless already set.

> Run every checkpoint marked ready in `docs/checkpoints.md`, not a hand-picked subset — the table is the list, and a new checkpoint should join the job by being marked ready, with no workflow edit.

Done, but with one honest carve-out. The script parses `docs/checkpoints.md`'s table for `**ready**` rows itself (no hard-coded checkpoint numbers), and separately asks `backend/scripts/checkpoint.py` which numbers *it* knows how to run — by invoking it with an invalid argument and reading the `(ready: 2, 3, 4, ...)` list off its own usage line, rather than duplicating that list here. It runs the intersection. Today that's checkpoints **2–9, 13–18, 20–23** (18 of them). The difference — **1** and the web checkpoints **10–12, 24–28** — is printed and skipped, not silently dropped: `docs/checkpoints.md`'s own "How a checkpoint is tested" section says these are walked by `make web-e2e` and by hand, not by `backend/scripts/checkpoint.py`; there is no HTTP-only backend runner for them to wire up. A new backend checkpoint joins this job the moment it's marked ready in the docs *and* added to `checkpoint.py`'s `CHECKPOINTS` dict — no edit here or in the workflow.

> Fail the job on the first failing step, and print the step's own words, not a stack trace.

Done. Each checkpoint runs as a subprocess (`python -m scripts.checkpoint N`) whose own ✓/✗ output streams straight to the job log unmodified; the runner stops at the first non-zero exit and does not wrap it. Confirmed below — the real run stops at checkpoint 2's own `✗ Pa reads his trail again (newest line is not Mei's NoKey: …)` line.

> It will be slow. Run it on pull requests against main and keep it out of the required set at first (strict=false, like image).

See "Timing" and "On `strict=false`" below — it isn't actually slow (~35s of checkpoint-walking once boot is included), but see the findings: it can't currently reach the end regardless.

> The checkpoints assume a clean database... Each checkpoint must pick a free port rather than a fixed one.

Done — throwaway sqlite file per run, in a `TemporaryDirectory` torn down in a `finally`; free port picked before boot. Verified: two consecutive local runs picked different ports (54730, then 57183, then 57575, then 60338) and left no sqlite file or orphaned `uvicorn`/`make` process behind afterward (checked with `lsof`/`ps` after each run).

> Do not let the job paper over a failure by retrying.

No retry logic anywhere in the script. One attempt, one verdict.

## The most important finding: several checkpoints are currently broken under a frozen clock — including a near-exact repeat of the #190 false report

I did **not** fix any of these — per the task, I'm reporting them for you to decide.

**Root cause, confirmed:** a frozen clock (as this CI job — and `make web-e2e` — must run with) means many events in one checkpoint's walk share the *identical* timestamp. At least one backend ordering query has no tie-breaker for that: `backend/app/audit/trail.py:147` is `order_by(AuditEntry.at.desc())` only. Under a real, ticking clock this is invisible (every write gets a distinct instant); under a frozen one, "newest first" becomes arbitrary whenever two-plus writes land in the same second — which, in a scripted walk that does a dozen actions in under a second, is most of them.

**Proof of the root cause**, not just the symptom: I ran the same 18 checkpoints, in the same order, against the same shared dev server, twice — once with `NURA_FROZEN_CLOCK` set, once without (checkpoint 20 was excluded from the unfrozen run since it explicitly requires a frozen clock and says so in its own refusal). Unfrozen, on a completely clean `dev.db`, freshly migrated from this branch, **all 17 pass** (checkpoint 20 separately passes when frozen, in ~4.6s). Frozen, **10 of 18 fail**, deterministically, on two independent full runs.

| # | Frozen clock | Real clock | What it says when it fails |
|---|---|---|---|
| 2 | **FAIL** | pass | `Pa reads his trail again (newest line is not Mei's NoKey: {..., 'action': 'write', 'target': 'profile', ...})` |
| 3–6 | pass | pass | |
| 7 | **FAIL** | pass | `Pa reads his audit trail (refusals on it: {'TooLongToSay', 'NotPlainEnough'})` |
| 8 | **FAIL** | pass | `Pa's app asks for the cached page (not the last first page rendered)` |
| 9 | **FAIL** | pass | `Pa reads the emergency lines of his trail (flag after ladder: [...])` |
| 13 | **FAIL** | pass | **`Mei reads Pa's notes as his chief (403, {"refusal":"OutOfScope","scope":"notes"}) — expected 200`** |
| 14–15 | pass | pass | |
| 16 | **FAIL** | pass | `Mei reads what changed again (expected only the new reading: {..., 'lines': [{'text': 'Nothing changed since Tuesday 15 September.', ...}]})` |
| 17 | **FAIL** | pass | `Pa reads his trail (a step is missing from the trail)` |
| 18 | **FAIL** | pass | `Pa reads his State again (not recomputed from the reading: {...})` |
| 20 | pass (needs frozen) | n/a (refuses cleanly: `the dev server's clock is not frozen: start it with NURA_FROZEN_CLOCK=... then run this again`) | |
| 21 | **FAIL** | pass | `Mei's metrics read is on Pa's trail (no nudge_metrics line)` |
| 22 | **FAIL** | pass | `Pa reads his feed (no logistics card the day before: ['now', 'visit', 'gate', 'story', 'story'])` |
| 23 | pass | pass | |

Checkpoint 13's failure — **word for word the same step named in #190**, "Mei reads Pa's notes as his chief" — is not a coincidence. I think this *is* what happened in #190: someone ran checkpoint 13 against a dev server that had a frozen clock left over (from a prior `make web-e2e` run, or set by hand), got this exact refusal, and reported it as a regression, when the real cause was the clock, not the code. That is the precise blind spot #192 was opened to close — and building the CI runner immediately reproduced it.

I have not diagnosed each of the other nine individually (7, 8, 9, 16, 17, 18, 21, 22) — they may not all share literally the same missing-tie-breaker cause as 2 and 13 (8 and 22 in particular smell like "the day before" / "cached page" logic keyed off a `now` that never moves, which is a related but distinct class of frozen-clock assumption), but all ten pass cleanly on a real clock and fail cleanly and deterministically on a frozen one. I'd treat this list as the actual todo, not the CI job itself.

**A design tension I want to flag rather than route around:** only checkpoint 20 documents needing a frozen clock (its own module docstring: `NURA_FROZEN_CLOCK=2026-09-14T06:00:00+08:00 make dev`); the other 17 backend checkpoints' docstrings just say `make dev`, and `docs/checkpoints.md`'s own "How a checkpoint is tested" section doesn't mention a frozen clock for them either. So it is arguable the *runner* is wrong to force `NURA_FROZEN_CLOCK` for the whole batch — except that #192 explicitly asks for it ("with the frozen clock"), one shared dev server is what makes an 18-checkpoint job fast rather than 18 separate boots, and checkpoint 20 needs *some* frozen instant to even start. I kept the frozen clock, on the read that #192 wants it and that the ten failures above are a real, valuable finding rather than an artifact of a wrong choice — but this is the one place I made a judgment call instead of just executing, and I'd rather you see it named than buried.

## Timing

The real job, run as designed, stops at checkpoint 2 in ~0.4–5s (machine-dependent) after a ~3–4s boot. To answer "how long would the whole set take" honestly, I also timed all 18 running to completion (the diagnostic table above): **~35s of pure checkpoint-walking time**, plus boot/migrate (~3–4s) — under a minute total. Once the ten failures above are triaged, this job will be fast; nothing about its timing argues against making it required eventually.

## On `strict=false`

Current branch protection (`gh api repos/ashfrnndz21/NuraApp/branches/main/protection`) requires `["backend", "backend-postgres", "image"]` — `image` **is** in the required set today, so "keep it out of the required set at first, like image" (from #192's text) doesn't quite match current reality; `web` is the job that's actually present in the workflow but absent from required checks. I'd suggest following `web`'s pattern: add the job to `ci.yml` normally (no `continue-on-error`, so it genuinely reports red/green) and simply don't add `checkpoints` to the branch-protection contexts list yet. That's a branch-protection-settings change, not something expressible in the workflow YAML — you'll need to do it in repo settings (or tell me to, once I'm allowed near `.github/`/settings).

## Gates

- `make lint` — clean (ruff + mypy, `264 source files`, `All checks passed!`).
- `make checkpoints-ci` — run repeatedly, see findings above; deterministic, not flaky.
- `make test` (full backend suite) — **deliberately not run to completion.** This diff touches no `backend/app` or `backend/tests` file (only `Makefile`, `.gitignore`, and the new top-level `scripts/run_checkpoints.py`, which is outside the backend package and untouched by `ruff check .`/`mypy app`'s backend-only scope), so there is no plausible regression for it to catch here; a partial run (26% in, all green) confirmed nothing was on fire. I stopped it early after noticing another builder's pytest run active on this shared Mac against a Postgres instance — running a second full suite in parallel would have contended with it for no benefit.
- `make plain-words` / `make language` — not run: no patient-facing string was added or changed anywhere in this diff (see the plain-words reviewer report below).
- No `make checkpoint N=<n>` gate applies in the story sense — this PR's whole content *is* the checkpoint runner.

### clinical-safety-reviewer (applied by hand — no profile/keys/consent/audit/medicine/reasoning/safety code in this diff to hand to the subagent)

1. Every profile read goes through the keys module — **N/A**, no profile-read code touched.
2. Every Fact carries provenance; every Card records its State — **N/A**.
3. No output starts/stops/changes a medicine — **N/A**.
4. Drug data from the licensed client only — **N/A**.
5. Red-flag paths escalate first — **N/A**.
6. High-risk drugs require a label photo — **N/A**.
7. Health data stays in region; none reaches analytics — **N/A**: the runner's throwaway sqlite database and object store live under the OS temp dir and are deleted at the end of every run; nothing is sent anywhere.
8. Consent recorded at claim and per key — **N/A**.

**Verdict: pass — nothing in this diff falls in scope.**

### plain-words-reviewer (applied by hand, same reason)

Every string this diff adds is developer/CI-facing only: `Makefile` comments, and the runner's own `print()` lines (ports, temp paths, pass/fail summary). None is rendered to a patient, caregiver, or family member, in any language, anywhere.

**Verdict: pass — no patient-facing string in this diff.**

- **No automated tests for `scripts/run_checkpoints.py` itself.** It's an orchestration script outside the backend package (like `scripts/backlog_to_issues.py`), and its own real behavior — boot, migrate, run, stop-on-first-failure, teardown — is exactly what four manual end-to-end runs above already exercised more meaningfully than a mocked unit test could. Happy to add a narrow test for just the markdown-table parser (`ready_checkpoints_in_docs`) if you'd rather have one on file.
- **No fix for the ten broken checkpoints.** Explicitly out of scope per the task.
- **No workflow edit.** Per the operator's instruction. YAML below.

## The exact YAML to add (`.github/workflows/ci.yml`, a new job under `jobs:`)

\`\`\`yaml
  checkpoints:
    # #192: walks every checkpoint marked ready in docs/checkpoints.md against a throwaway
    # dev server this job boots, migrates and tears down itself (backend/scripts/
    # run_checkpoints.py; make checkpoints-ci) — its own free port, its own sqlite file, no
    # state shared with any other job. The closest thing this repo has to "does the product
    # actually work end to end," and the only CI check that would have caught the class of
    # gap #190 found by hand.
    #
    # Not in branch protection's required checks yet (see the PR that added this job): on
    # main as of 2026-09-17, several checkpoints fail under this job's frozen clock (a
    # missing tie-breaker on tied timestamps, not a flake — reproduces every run) and need
    # triage first. Expected to go red; that is the job doing its job.
    runs-on: ubuntu-latest
    timeout-minutes: 15
    steps:
      - uses: actions/checkout@v4
      - uses: actions/setup-python@v5
        with: { python-version: "3.12" }
      - run: pip install -e "backend[dev]"
      - run: make checkpoints-ci
\`\`\`

No `continue-on-error`, no changes to `required_status_checks` — it reports real pass/fail on every PR against `main`, same as `web` does today, and stays optional until branch protection is updated separately to require it.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


ashfrnndz21:
**Operator merge note.**

Merging the runner. **Not** adding the CI job yet, deliberately.

`scripts/run_checkpoints.py` walks every checkpoint marked `ready` in `docs/checkpoints.md`. The table is the list, so a new checkpoint joins by being marked ready, with no workflow edit. It boots the app on a **throwaway database it creates and destroys**, a **free OS-assigned port** and a frozen clock, stops at the first failing checkpoint, and prints that checkpoint's own words rather than a stack trace. No retries. `make checkpoints-ci` runs it.

**What building it found is the reason this PR matters.** With the clock frozen, **10 of the 18 runnable checkpoints fail, deterministically**; without it they pass. The cause: queries that pick "the newest" by timestamp alone, with no tie-breaker. When several rows share an instant, which one counts as newest is arbitrary. Checkpoint 13's failure, "Mei reads Pa's notes as his chief", is word for word the report #190 could not reproduce. That report was a real bug, not a dirty worktree, and I have corrected the record on #190.

It reaches past the audit trail: 16 queries order this way, including ones that decide current state (which delivery settings apply, which login challenge is valid, which red flag is current, which "which tablet?" question is open). That fix is being built now, `risk:high`.

**Why the job waits:** added today, it would fail on every run. A check that is always red teaches everyone to ignore it, and then it misses the real failure it was built to catch. It goes in once the ordering fix lands and the 18 pass with a frozen clock, several runs in a row. It will start non-required so we can learn how flaky it is before it can block a merge.

The runner also adds `.venv-ci/` and `.mypy_cache/` to `.gitignore`. A blanket `git add` in this worktree earlier swept 4,179 cache files into a commit.

Issue #16, opened 2026-09-13
E02-04 WhatsApp inbound ingestion: forward a photo or PDF to the app number and it file


ashfrnndz21:
**Epic** E02 Capture and ingestion
**Persona** All · **Tier** T1 · **Priority** P1 · **Size** M

**Story**
WhatsApp inbound ingestion: forward a photo or PDF to the app number and it files itself

**Acceptance**
Forwarded item appears in the record with a confirmation reply within a minute

**Depends on**
E02-01,E11-01

**Spec** see docs/00-MASTER-BUILD-SPEC.md section 0 for the document that covers this epic.

ashfrnndz21:
Delivered in #107 (E19 WhatsApp agent), merged to main.

Issue #15, opened 2026-09-13
E02-03 PDF import from portals, email and shares


ashfrnndz21:
**Epic** E02 Capture and ingestion
**Persona** Chief · **Tier** T1 · **Priority** P1 · **Size** S

**Story**
PDF import from portals, email and shares

**Acceptance**
Multi-page PDFs parsed into events with dates

**Depends on**
E02-01

**Spec** see docs/00-MASTER-BUILD-SPEC.md section 0 for the document that covers this epic.

ashfrnndz21:
Delivered in #113 (E02 capture), merged to main.

PR #228, opened 2026-09-17
A real monotonic tiebreaker for every "newest wins" read, not a frozen clock that ticks (#192/#218)

ashfrnndz21:
Follow-up to #218: its checkpoint runner found ten checkpoints failing, deterministically, only under its frozen clock — the sixteen "newest wins" queries the coordinator named, each ordering by a timestamp column alone. This PR classifies and fixes all sixteen, in one place (`app.db.monotonic`), plus one more the sixteen didn't cover that was still blocking checkpoint 13 after the sixteen were fixed. `risk:high`: this touches audit, consent, keys-adjacent code, a migration, and delivery/safety reads.

## What was asked, answered

> Classify them as "latest wins" decisions (need true insertion order) versus display ordering (needs determinism), fix all 16 in one risk:high PR, and don't make the frozen clock tick.

Done — table below. Nothing makes the clock tick; the fix is a real database-issued sequence number (`seq`), independent of the clock entirely.

> Done means all 18 runnable checkpoints pass with a frozen clock, several runs in a row.

**Not fully met, and I want to be precise about why rather than call it done.** All sixteen are fixed, and fixing them plus the seventeenth (`Consent`, below) gets checkpoint 13 *past* the exact step named in #190 and #192 ("Mei reads Pa's notes as his chief") — confirmed twice, deterministically, against a fresh throwaway db each time. But checkpoint 13 still fails one step later, and three other checkpoints (16, 18, 22) fail too, on **three different bugs that are not this one** — none of them an unordered "newest" read. I traced each far enough to be confident of that classification (detail below) and stopped, per the instruction not to fix a different bug in this PR. 14 of 18 pass.

## The sixteen, classified

| # | Site | Table | Kind | Fix |
|---|------|-------|------|-----|
| 1 | `audit/trail.py:147` | `AuditEntry` | **Decision** (see note) | `.seq.desc()` |
| 2-4 | `delivery/triggers/deliver.py:360`, `preferences.py:45,64` | `DeliverySettings` | Decision — which settings apply | `.seq.desc()` |
| 5 | `safety/red_flags.py:1132` | `Flag` | Display — every flag processed regardless of order | `.id` |
| 6 | `identity/login.py:86` | `LoginChallenge` | Decision — `open_challenges[0]` is checked against the code | `.seq.desc()` |
| 7 | `channels/whatsapp/inbound.py:1100` | `DoseQuestion` | Decision — which tablet a reply answers | `.seq.desc()` |
| 8 | `memory/changes.py:169` | `LastLooked` | Decision — the diff baseline | `.seq.desc()` |
| 9 | `family/thread.py:128` | `ThreadMessage` | Decision (table-wide, see note) | `.seq.desc()` |
| 10 | `reasoning/feelings/service.py:429` | `FeelingNote` | Decision (table-wide, see note) | `.seq.desc()` |
| 11 | `reasoning/feelings/record.py:190` | `FeelingTap` | Decision — feeds `last_tap_at`, a gate read elsewhere | `.seq.desc()` |
| 12 | `safety/emergency_card.py:307` | `Event` | Decision — the latest reading on the card | `.seq.desc()` |
| 13 | `delivery/feed/search.py:228` | `SearchJob` | Display — a listing, every job shown | `.id` |
| 14 | `delivery/feed/rank.py:498` | `FeedPage` | Decision — which page is "the" offline cache | `.seq.desc()` |
| 15 | `delivery/nudges/engine.py:345` | `FeelingNote` | Decision — returns on the first match in the loop | `.seq.desc()` |
| 16 | `delivery/nudges/engine.py:454` | `ThreadMessage` | Decision — returns on the first match in the loop | `.seq.desc()` |

**Two corrections to my first pass, both caught by actually reading each caller rather than trusting the query's shape:**
- **#9 and #10 (`ThreadMessage`, `FeelingNote`) are each read from two places** — one a plain listing (would have been "display" alone), the other (#15, #16) an early-`return` inside a `for` loop, which makes it a decision. Since both reads are the same table, both get `.seq` — no reason for two tiebreak strategies on one table.
- **#1 (`AuditEntry`) is not a plain listing.** I first classified it "display" (every touch is written, order is cosmetic) and was wrong: checkpoint 2 immediately caught it — `AuditEntry.id` as tiebreak deterministically reproduced a *different* wrong "newest" line than the missing-tiebreak version, because the trail's entire purpose is telling him, truthfully, what happened and in what order. A stable-but-arbitrary tiebreaker is not good enough for a record whose job is being trusted; it needed `.seq`, the real one. Fixed before this PR's first push.

## The seventeenth: `Consent`, found while proving the sixteen fixed checkpoint 13

Fixing all sixteen still left checkpoint 13 failing on **"Mei reads Pa's notes as his chief (403, OutOfScope, notes) — expected 200"** — word for word the false regression #190 reported and could not reproduce, and #192 was opened over. Traced it: `app.consent.service.require_consent` (the gate a key is cut from) picks the applicable consent in **Python**, not SQL —

```python
for row in sorted(active, key=lambda row: as_utc(row.granted_at), reverse=True):
    if row.text_version == wanted:
        return ConsentCheck(..., scopes=...)
```

— which a `grep -n '.order_by('` for the sixteen never finds. In checkpoint 13, Pa gives Mei two `SHARE_WITH_PERSON` consents moments apart (the claim's, then one naming her notes too); under the frozen clock both carry the identical `granted_at`, tie, and the tie let the *older, narrower* one win at random. The chief key cut on it opened everything but the notes it was meant to add. `Consent` gets `seq`, and the sort (here, plus the two "oldest first" listings, `active_consents`/`all_consents`) is `(granted_at, seq)`.

I want to flag the safety direction of this one explicitly for the reviewer: the tie could just as easily have resolved the other way — an older, *wider* consent winning over a newer, narrower one (say, after Pa **narrows** what Mei may see), which would have **over-granted** a key rather than under-granting it. Checkpoint 13 only ever exercised the direction that fails safe. This was a real, silent, pre-existing risk in consent resolution; the fix removes it in both directions, not just the one the checkpoint happened to catch.

With both fixes: checkpoint 13 gets *past* this exact step, confirmed on two separate fresh-throwaway-db runs.

## A wider sweep, reported and not fixed

`grep -rn granted_at app/` turned up about a dozen more places that sort `Key` rows by `granted_at` with no tiebreaker, or a weak one (`family/reorder.py` and `safety/not_feeling_well.py` already carry an ad hoc `str(id)` tiebreak of their own — prior art for exactly this problem). I checked each for the "returns on first match" shape that made #9/#10 and the Consent read real bugs rather than cosmetics:

- **All but one** (`delivery/reach.py`, `delivery/triggers/deliver.py:289`, `delivery/feed/compose.py:170`, `channels/whatsapp/group.py:108`, `channels/whatsapp/inbound.py:244`, `channels/whatsapp/outbound/level0.py:350`, `channels/api/profiles.py:682`, `family/grants.py:126`, `state/dimensions.py:294`, `safety/emergency_card.py:279`, `memory/changes.py:622`, `identity/doors.py:586`, `safety/red_flags.py:1156`) only build a list or a line of prose from **every** key in the loop — the same shape as the trail or the self-search queue (display, not a decision).
- **One** (`channels/whatsapp/outbound/level0.py:323`) does `for key in sorted(...): if ... role is CHIEF: return person.display_name` — an early return, the same shape as the bugs I did fix. I did **not** fix it: I have no checkpoint or test proving it currently picks the wrong chief (a profile would need two simultaneously-active chief keys, which I have not constructed), so unlike `Consent` this is reported, not proven, and I'd rather flag a suspect than "fix" something I can't demonstrate is broken.

None of these is touched in this PR.

## The other three checkpoint failures — different bugs, not fixed here

Per the instruction, traced each far enough to classify it, then stopped.

**Checkpoint 13** (after the consent fix), **step**: `Kit reads his digest` — expects `"It was 138 over 84."` in the lines, gets `['Pa on Monday 14 September.', 'There is nothing new since Wednesday 16 September.', 'Mei is on duty today.']`. Root cause: `backend/scripts/checkpoints/cp13.py:547` computes `since = (datetime.now(UTC) - timedelta(days=1)).isoformat()` — the **real wall clock**, not the frozen app clock (`2026-09-14`). This machine's real date is 2026-09-17, so `since` lands on 2026-09-16 — a moment *after* the frozen app's "now", so nothing on the profile is ever new enough. This is a bug in the checkpoint script itself, not the product, and it gets worse every real day that passes; not an ordering bug, so not touched here.

**Checkpoint 16**, **step**: `Mei reads what changed again` — expects the new blood-pressure reading in the lines, gets "Nothing changed since Monday 14 September" with `looked_at` exactly equal to the new fact's timestamp. Traced to `app/memory/changes.py`'s `newer(column) = column > after` — a **strict** `>`. Under the frozen clock, the look and the new fact share the identical instant, so the new fact is not "newer" by a strict inequality, even though it happened after the look in write order. This needs `since` to become a `(timestamp, seq)` cursor, not just a timestamp — a change to the `what_changed` API's own contract, materially bigger than adding a tiebreaker to an existing `ORDER BY`, so not attempted here.

**Checkpoint 18**, **step**: `Pa reads his State again` — expects a fresh recompute from a second reading, gets the prior snapshot back. Same shape as checkpoint 16 by inspection (a `>` boundary against `computed_at` under a clock that never advances) but I did not trace it as far as 16; reporting it as the same suspected class, not confirmed.

**Checkpoint 22**, **step**: `Pa reads his feed` — expects a visit-logistics card "the day before" a visit and doesn't get one (`['now', 'visit', 'gate', 'story', 'story']`). I have not traced this one; it may be the same `>`-boundary class or a genuinely separate date computation. Not fixed.

## Checkpoint results (all 18 runnable, fresh throwaway db + free port + frozen clock per run — never a shared server, on the coordinator's note that a reused one is exactly the #190 contamination shape)

| # | Result |
|---|---|
| 2–9 | **PASS** (8/8) |
| 13 | FAIL — different bug, above |
| 14 | **PASS** |
| 15 | **PASS** |
| 16 | FAIL — different bug, above |
| 17 | **PASS** |
| 18 | FAIL — different bug, above (unconfirmed root cause) |
| 20 | **PASS** |
| 21 | **PASS** |
| 22 | FAIL — different bug, above (untraced) |
| 23 | **PASS** |

**14 of 18 pass.** Each run booted its own `make dev`, migrated a fresh sqlite file in a temp directory, picked a free port, and tore both down after — the same mechanics `scripts/run_checkpoints.py` (#218) uses, run per-checkpoint here rather than as one shared boot, so no checkpoint's result depends on another's.

## The migration (`0040_monotonic_tiebreak`)

`seq_counters` (one row per sequenced table) plus a `seq` `BigInteger NOT NULL` column (server_default `0` — this app has no production data yet to reorder properly) on twelve tables: the eleven decision tables above plus `consent`. `app.db.monotonic` is a class decorator; a `before_insert` mapper event does one atomic upsert (`INSERT ... ON CONFLICT (name) DO UPDATE SET value = value + 1 RETURNING value`) — portable across SQLite (serialised for free: every write already begins IMMEDIATE) and Postgres (the counter row's own lock), and I checked empirically that SQLAlchemy's `Identity()`/autoincrement does **not** work for a non-primary-key column on SQLite before choosing this over that more obvious-looking route.

`seq_counters` is a plain Core `Table` on `Base.metadata`, not an ORM model — needed because `tests/conftest.py` builds the test database from `Base.metadata.create_all`, not by running migrations; without this every test creating a row of any of the twelve tables would fail with "no such table" inside the insert event.

## Gates

- `python3 -m ruff check .` and `python3 -m mypy app` — clean, 264 source files.
- `python3 -m pytest -q tests/test_data_map.py tests/test_migration.py` — 8 passed (migration builds exactly what the models declare, both ways; downgrades to nothing).
- Targeted subset for every module this diff touches (no full suite — shared machine, per the "Hard-won rules"): `test_audit_acceptance.py test_identity.py test_triggers.py test_whatsapp_taken_which.py test_memory.py test_emergency_card.py test_feed.py test_feelings_cloud.py test_feelings_inference.py test_family_thread.py test_nudges.py test_nudge_metrics.py test_red_flags.py` (200 passed) and `test_consent.py test_consent_acceptance.py test_consent_decoupled.py test_consent_withdraw_api.py test_family_group_consent.py test_family_grants.py test_family_roster.py test_family_privacy.py` (83 passed). 283 passed total, 0 failed — CI's `backend`/`backend-postgres` are the full-suite gate.
- `make plain-words` / `make language` — not run: no patient-facing string in this diff (see plain-words review below).
- Checkpoints: table above.

### clinical-safety-reviewer (applied by hand against `.claude/agents/clinical-safety-reviewer.md`'s eight checks)

1. Every profile read goes through the keys module with a key context — **pass**. No new read path; every changed query already went through `audited_read`/`scoped_select`. The one genuinely new read (`seq_counters`, inside `monotonic`'s insert hook) touches no profile data — it's a table name and a counter, classified `operational` in the data map.
2. Every Fact carries provenance; every Card records its State — **N/A**, untouched.
3. No output starts/stops/changes a medicine — **N/A**, untouched.
4. Drug data from the licensed client only — **N/A**, untouched.
5. Red-flag paths escalate before ranking/caps — **pass, checked directly**. `red_flags.py`'s `open_flags` ordering changed (added `.id` tiebreak); the three callers that climb ladders or build cards (`triggers/engine.py`, `triggers/day.py`, `feed/compose.py`) all iterate *every* flag in the result unconditionally — confirmed by reading each, no early return — so escalation still happens for every flag regardless of the (now merely deterministic) order they're read in.
6. High-risk drugs need a label photo before a dose is saved — **N/A**, untouched.
7. Health data stays in region; none reaches analytics — **pass**. `seq_counters` and every `seq` column hold no health content; nothing new is sent anywhere.
8. Consent recorded at claim and per key; refusals visible in the audit — **pass, and worth reading closely**: the `Consent` fix is a genuine safety improvement, not just a checkpoint fix. Before it, a tied `granted_at` could resolve to *either* the older or the newer matching consent at random — checkpoint 13 only exercises the direction that fails safe (narrower wins), but the same bug could just as easily have let an older, *wider* consent win after Pa narrowed what someone may see, over-granting a key silently. The refusal path below the picking loop (`ConsentOutOfDate`/`ConsentRevoked`/`NoConsent`) is untouched — only which already-valid consent is picked changed.

**Verdict: pass, no required changes.** One item to flag for judgement, not a blocker: `channels/whatsapp/outbound/level0.py:323` has the same "early return on first match over an unordered `sorted(keys, ...)`" shape as the bugs fixed here (picks a chief's display name to send an outbound message under), but I found no checkpoint or test proving it currently picks the wrong chief, so I reported rather than fixed it (above).

### plain-words-reviewer

**Verdict: pass, nothing in scope.** Every string this diff touches is developer-facing: code comments, docstrings, the migration's own prose, `scripts/data_map`'s classification labels (`operational`). No patient, caregiver, or family-facing string was added or changed anywhere in this diff.

- **The dozen `Key.granted_at` display-ordering sites** (listed above) — same shallow risk as the trail/queue class already left with `.id`-only leveling elsewhere in the codebase; not proven to break anything, and fixing a dozen files on speculation felt like scope creep against "fix the sixteen, report the rest."
- **`channels/whatsapp/outbound/level0.py:323`** — same reasoning, flagged above for the reviewer specifically since it does share the exact early-return shape of the confirmed bugs.
- **Checkpoints 13 (final step), 16, 18, 22** — three to four distinct, unrelated bugs, none an ordering-tiebreak problem; each traced and reported above rather than fixed, per the instruction.
- **A historical backfill for `seq=0` on any pre-existing row** — this app has no production data yet; a fresh row always sorts after a migrated one, which is the safe direction to be wrong in.

## Branch note

Branched from current `main` (`6f090a57`, which already includes #218 merged), not from #218's now-closed branch, per the coordinator's note. `scripts/run_checkpoints.py` and its `Makefile` target came in with that merge; this PR does not touch either.


ashfrnndz21:
## Fixed: the mechanism, per independent review

The review is right, and reproduced it rather than just reading it — replaced now.

**What was wrong:** `seq_counters`'s `UPDATE ... RETURNING` upsert takes a Postgres row lock held until the transaction that ran it commits. This app runs one transaction per HTTP request, and `app.audit.trail.record` writes an `audit_entry` — one of the twelve `@monotonic` tables — on nearly every audited read and write. So the `audit_entry` counter row was locked from a request's first audit line until the whole request finished, and every other request queued behind it; two transactions taking two tables' counter rows in opposite order deadlocked outright.

**The fix, exactly as prescribed:** on Postgres, `seq` now comes from a real `SEQUENCE` (`<table>_seq_seq`) via `nextval()` — not transactional, takes no row lock held for the transaction's life, cannot block a concurrent insert, cannot deadlock. `app.db.monotonic` decides this internally (dialect-checked in the one `before_insert` hook); no caller, `order_by` site, or classification changes. SQLite (dev/tests only, already serialises every write via `BEGIN IMMEDIATE`) keeps the original `seq_counters` row — verified that's still necessary and sufficient there, not just assumed.

**Verified against a real, local Postgres 16** (initdb'd fresh on this machine, not the CI service — wanted hands-on proof before trusting it to a second review):
- `SELECT nextval(...)` from an uncommitted second connection returns in ~2ms while a first connection's own uncommitted `nextval()` stays open — the review's exact experiment, this time not blocking.
- `alembic upgrade heads` / `downgrade base` / `upgrade heads` / `check` all clean on that Postgres — the migration's new `CREATE SEQUENCE` (guarded to Postgres only) builds a schema `alembic check` reports no drift against.
- `tests/test_migration.py` + `tests/test_data_map.py` pass on both SQLite and that Postgres.
- The previous commit's 283-test subset still passes, unchanged, on SQLite.
- `scripts/run_checkpoints.py` still passes checkpoints 2–9 and stops at 13 the same way as before this commit (the separately-reported, unrelated bug) — the mechanism swap is invisible to the product.

**New:** `backend/tests/test_monotonic_seq_concurrency.py`, Postgres-only (skipped everywhere else, like the existing `test_confirm_race.py` it's modelled on; runs in `backend-postgres` CI). Two tests, 1.35s total against my local Postgres:
1. two sessions each write an `audit_entry`, the first's transaction left open — the second's write must not be slowed by it (asserted under 1s; actually ~2ms).
2. two sessions each touch `audit_entry` then `consent` in opposite order, run genuinely concurrently (`asyncio.gather`) — must not deadlock.

**Also done:** the migration's docstring now states `seq=0` on existing rows assumes no production data exists yet to reorder properly — an assumption, not a verified fact, stated as such rather than settled.

**Unchanged, confirmed still correct:** all 11 decision tables' append-only property, the `Flag`/`SearchJob` display-only classification, the consent fix's correctness, the data map, and checkpoint 16's separate `>`-boundary cause — none of this PR's classification or the four separately-reported checkpoint failures needed to move.

Pushed to `audit-monotonic-tiebreak` (and `wip/audit-monotonic-tiebreak`) at `c67b00d5`. `git diff 5104f22a..c67b00d5 -- backend/app/db.py backend/migrations/versions/0040_monotonic_tiebreak.py` is the whole mechanism change.

PR #229, opened 2026-09-17
A medicine is never withheld from the emergency card again, in every language (#222)

### LINKED ISSUE #222: A medicine can silently vanish from the emergency card when the register has no plain name for it
## What happens

A medicine he takes can **silently disappear from the sentences on his emergency card**.

Found while testing #171 (PR #221) with paracetamol, and traced in the code:

1. `_plain_medicine` (`backend/app/safety/emergency_card.py:344-352`) gives a medicine his own name for it, "the water pill", from the register's story. When the register has **no story** for that generic, it falls back to the register's name: `line.generic.title()`, so "Paracetamol".
2. That sentence goes through `render`, which runs the plain-words verifier. A medical term not in the glossary fails it with `NotPlainWords`.
3. On `NotPlainWords`, the card **withholds the line and logs a warning** (`emergency_card.py:404-409`). No error or placeholder appears, and nothing tells him or his family.

The medicine is still in the card's structured data, but **the sentence naming it is gone.**

## Why this is serious, and why now

The emergency card is what a stranger reads when he cannot speak for himself: a paramedic, an A&E nurse, a neighbour. A medicine missing from it is how someone is given a drug that interacts with one he is already on, or how a bleed on an anticoagulant goes unrecognised.

It fails silently, and a silent gap is the worst kind on a safety artefact: the card looks complete.

**It just got much more likely.** #202 grew the drug register from 17 to 233 products, and most of them have no plain-name story. So on current `main`, many of his medicines may be affected, not just the odd one.

## The fix, and the principle behind it

`emergency_card.py:132-133` already states the principle: *"`strength` and `generic` are the register's words, data for the stranger; `plain_name`, `amount` and `when` are his."* The register's name is **data for the stranger**, not a sentence to him. A paramedic needs "Warfarin", not a plain paraphrase of it.

So:
- **A medicine must never be withheld from the emergency card.** Every active medicine appears, always.
- Where there is no plain name, the card still names it by the register's name. The plain-words check must not block a clinical name on the one artefact written for a clinician. That may mean the register's name is carried as data rather than through the verifier, or a fallback sentence built to pass it. Choose deliberately and write down why.
- Plain words still bind on everything that *is* addressed to him.
- **A withheld line on the emergency card must never be silent again.** If any line still cannot render, it must surface somewhere a person will see, not only in a log.

## Tests
- **Every active medicine appears on the emergency card**, including one with no plain-name story, in all three languages. Assert over the whole register, not one hand-picked drug, so a product added later cannot fall outside the check. That is the #186 lesson: a hand-picked list is the bug.
- A high-risk medicine with no story still appears, with its high-risk marker.
- The printable page and the offline-kept card show the same medicines as the live card.

ashfrnndz21:
Closes #222

## The measurement (first, as asked)

Before changing anything, I measured how many of the register's medicines vanish from the
emergency card on current `main` (`2c30a3c5`), over the **whole** fixture register
(`backend/tests/fixtures/drugs/registry.json`, 50 generics / 233 products), in each language:

| language | medicines that vanish | of |
|---|---|---|
| en | **26** | 50 (52%) |
| ms | 0 | 50 |
| zh | 0 | 50 |

So on `main` today, more than half of every English-language emergency card's medicines that
have no glossary-covered chemical name are silently missing their naming sentence — paracetamol
included, exactly as found in #171/#221. Malay and Chinese are unaffected because their
sentences never carry the register's chemical name (see mechanism below).

This is reproducible: `pytest -q backend/tests/test_emergency_card.py::test_every_active_medicine_on_the_whole_register_appears_on_the_card_in_every_language`
now asserts 0/50/0 in all three languages; I confirmed the 26/0/0 split by loading the pristine
`main` copy of `_plain_medicine`/`render` (`git show 2c30a3c5:...`) and running it over the same
registry before touching any code.

## The question I was asked to answer, first: is the chemical name visibly shown on every surface a stranger reads?

**Before this round: no.** The live phone screen (`web/src/screens/Emergency.tsx`) rendered
only `card.lines.map(line => line.text)` — the verified sentences — and nothing else; it never
touched `card.medicines[].generic` or `.strength`. Being in the JSON was not being on the
screen. For English, that meant the chemical name was invisible on the phone for the same 26 of
50 generics whose sentence can't carry it (paracetamol: "Pa takes the pain tablet." — no
"paracetamol" anywhere on screen); for a high-risk drug with a non-enrichable name — methotrexate
in this fixture — the `ec.high_risk` marker fired ("Pa's doctor watches the joint tablet
closely.") with no way to tell *which* high-risk drug it meant. For Malay and Chinese, it meant
the chemical name was invisible for all 50, always, sentence or no sentence — pre-existing, but
on the same screen. The offline-kept copy is the same JSON, so it inherited the same gap. It
was also **invisible as a failure**, not just invisible as data: `_medicine_label` catches
`NotPlainWords` itself, so `say()` never sees anything go wrong, `withheld` stays empty, and
`ec.render_issue` never fires — the original #222 fix moved "the line vanishes" to "the chemical
name vanishes, silently," on the one screen a paramedic actually reads.

**Now: yes**, on every surface, every key role, every language. `Emergency.tsx`'s `EmergencyCard`
component gained a data section, rendered straight from `card.medicines` — not through any
sentence, so plain-words does not apply to it, per the principle `emergency_card.py:132-133`
already states for `strength` and `generic` — showing, per active medicine: the plain name, the
chemical name, the strength, and a high-risk marker where it applies. One component, so it's the
live screen and the offline-kept copy at once (both read the same `KeptCard`), for the owner, a
chief and an emergency-only key alike (`her density`/`a neighbour's key` tests below all exercise
the same component). The chemical name and strength are the register's own words, so they render
identically in English, Malay and Chinese — verified by e2e test and by hand (see gates and
screenshot below).

Screenshot taken against my own dev server (potassium chloride, exactly the example asked for):
the sentence reads "Pa takes the body salt tablet (potassium chloride)." and, further down the
same card, the new data line reads "the body salt tablet — potassium chloride, 600 mg" — wraps
onto two lines cleanly at 375px width (mobile viewport), no clipping, sitting in the card's
existing caption style, right above the call/print buttons.

## The actual mechanism (slightly different from the issue's literal description)

The issue describes the bug as: no plain-name story → `_plain_medicine` falls back to
`line.generic.title()` → that fails the plain-words check. I measured that path too, and it
turns out **every one of the fixture's 50 generics already has a plain-name story** — so that
exact fallback never fires against this fixture today. What actually fires, for 26 of the 50:

1. `_plain_medicine` (old code) appended the register's chemical name in English regardless:
   `f"{plain} ({line.generic})"`, e.g. `"the water pill (frusemide)"`.
2. That string goes through `render()`'s plain-words verifier. Rule 3 (short words) exempts a
   chemical name from the "long word he'd have to ask about" check **only** when it's one of
   the handful of names `app.safety.plain_words.GLOSSARY` already lists (furosemide/frusemide,
   amlodipine, atorvastatin, metformin — the four pre-#202 drugs the register used to be built
   around).
3. #202 grew the register from 17 to 233 products across 50 generics. None of the 46 new
   generics (bisoprolol, paracetamol, amoxicillin, …) are in that four-name glossary, so their
   parenthetical fails rule 3, and the whole `ec.medicine` sentence — and, for a high-risk drug,
   the `ec.high_risk` marker line too, since it reuses the same label — is silently withheld:
   `emergency_card.py`'s old `say()` just logs a warning and drops the line.

The true "register has literally no story" case (`UnknownDrug`/no `plain_name_id`) does happen
to already pass today, because the fallback's title-cased name (`"Warfarin"`) is read by rule 3
as a proper name (any capitalized, non-first word is skipped) — the same mechanism the
verifier already uses for doctor/chief/insurer names. I kept and documented that as the
deliberate answer for that case (see below); it isn't new, and I didn't touch it to make it
"work" — it already worked, I just made sure it's what's used, and added a test asserting it
for a synthetic no-story, high-risk generic.

## The fix

**No active medicine is ever withheld, by construction, in every language:**

- `_plain_medicine` (`backend/app/safety/emergency_card.py`) no longer appends the register's
  name to the sentence at all — it returns the plain phrase alone (or the register's own name,
  title-cased, when there is no story), plus a new `has_plain_name: bool` saying which.
- `_medicine_label` (new) decides, per medicine, whether the register's name can safely stand
  beside the plain name in the **English** sentence: it tries `render()` with the parenthetical
  and falls back to the plain name alone if that fails. The parenthetical is now opportunistic
  — a bonus when the chemical name happens to be short/glossary-known — never load-bearing.
  `ec.medicine` and `ec.high_risk` both use this same label, so they can never raise for a
  medicine's own name.
- The register's own name is **always** on the card as data regardless of what the sentence
  says, via the existing `Medicine.generic` field (JSON) and now also in the printable page's
  data table (`printable.py`), which previously only showed `plain_name` — for a drug with no
  story it already equals `generic`, so the caption is added only when `has_plain_name` is true
  and would otherwise be missing. This matches the module's own stated principle
  (`emergency_card.py:132-133`, unchanged): `strength`/`generic` are the register's words, data
  for the stranger; `plain_name` is his. I chose "carry as data, only opportunistically as
  sentence" over "weaken the verifier" or "grow the glossary" specifically because the issue
  ruled those two out, and because the printable page and the JSON already had a data channel
  for exactly this (strength, phone numbers, policy references all already go this way) — the
  bug was that the chemical name was going through the verified sentence at all, not that the
  verifier was wrong to reject it there.

**A withheld line is never silent again:** `compose_lines`'s `say()` helper now remembers every
template id it had to drop; if any (non-medicine) line still fails for an unrelated reason, two
new static, pre-verified lines are appended to the card — `ec.render_issue` and
`ec.render_issue_family` — in `backend/app/channels/safety_strings.py`, all three languages, and
printed on the printable page too. They're static by design (no clinical content, nothing but
`{name}`, a slot already used elsewhere on the card) so they cannot themselves fail the standard
and go silent a second time. **Wording was revised once already, from the plain-words review
below** — see "Fixed during review."

**High-risk marker:** unaffected mechanically — `ec.high_risk` already reused whatever label
`ec.medicine` used, so it inherits the same guarantee. New test:
`test_a_high_risk_medicine_with_no_story_still_appears_with_its_marker` (synthetic
`oxymorphone`, no monograph in the fixture, `high_risk=True`) — appears, marker included, all
three languages.

## Why the printable page and the offline-kept card show the same medicines as the live card

I read `web/src/offline/emergencyCache.ts` (touched recently by #213/#221) and
`web/src/screens/Emergency.tsx` in full before deciding whether they needed changes. They
don't: the offline cache stores the backend's JSON (`card.lines`) and printable HTML verbatim,
and the live screen (`Emergency.tsx`) renders `card.lines` directly with **no independent
fallback to `card.medicines`** — so on the live screen, a withheld sentence used to mean the
medicine was genuinely invisible, not just missing a caption. Since the fix guarantees
`card.lines` always carries every medicine's sentence, the live screen, the printable page and
the offline-kept copy are fixed together, from one place, with no web changes needed beyond a
type addition (`has_plain_name` on `web/src/api/types.ts`, kept in sync with the new backend
field).

## Tests

- `test_every_active_medicine_on_the_whole_register_appears_on_the_card_in_every_language` —
  iterates `REGISTRY.generics` (not a hand-picked list — the #186/#215/#222 lesson the shared
  brief calls out), asserts every one's `ec.medicine` line renders, in en/ms/zh, and asserts the
  safety-net line never had to fire.
- `test_a_high_risk_medicine_with_no_story_still_appears_with_its_marker` — synthetic
  no-monograph, high-risk generic; asserts it's named and carries `ec.high_risk`, all three
  languages.
- `test_the_printable_page_and_the_live_card_show_the_same_medicines` — real end-to-end profile
  (frusemide + bisoprolol, the latter a real #202-era generic that used to vanish); asserts the
  live card's lines, and separately the printable page's sentences *and* its data table, all
  name both medicines, with bisoprolol's register name present as data on the page even though
  its sentence can't carry it.
- Updated `test_emergency_card.py`'s existing assertion that `medicine.plain_name ==
  "the water pill (frusemide)"` to `medicine.plain_name == "the water pill"` (data field,
  now pure) plus `has_plain_name`, since the parenthetical moved from data into a
  sentence-composition decision — the sentence text itself (`"Pa takes the water pill
  (frusemide)."`) is unchanged and still asserted.

## Acceptance criteria, answered

- **Every active medicine appears on the emergency card, asserted over the entire register, in
  all three languages** — yes, `test_every_active_medicine_on_the_whole_register_appears_on_the_card_in_every_language`.
- **A high-risk medicine with no story appears with its marker** — yes,
  `test_a_high_risk_medicine_with_no_story_still_appears_with_its_marker`.
- **The printable page and the offline-kept card show the same medicines as the live card** —
  yes; see above and `test_the_printable_page_and_the_live_card_show_the_same_medicines`. The
  offline cache needed no code change (verified by reading it in full).

## Gates (targeted files only, per the shared brief's hard-won rules — never the full suite on
this shared machine)

Re-run after every round below (the wording fix, the clinical-safety fixes, and the
independent-reviewer fix); numbers here are the final ones:

- `pytest -q backend/tests/test_emergency_card.py backend/tests/test_emergency_card_insurer.py backend/tests/test_plain_words.py` → **193 passed**
- (first round only) also ran with `test_drug_registry_conformance.py`, `test_not_feeling_well.py`, `test_safety_api.py` → **217 passed**
- `ruff check backend/app/safety/emergency_card.py backend/app/channels/safety_strings.py backend/app/channels/printable.py backend/app/channels/api/safety_schemas.py backend/tests/test_emergency_card.py` → **all checks passed**
- `mypy` on the same backend files → **no issues found**
- `make plain-words` → **8034 strings, 0 failures** (120 pre-existing notes, 1 pre-existing exempt) — unchanged by the medicine-data section, deliberately: that text is not `@patient` tagged (see above)
- `make language` → **8173 strings, 0 failures** (89 pre-existing notes)
- `make build-web` (`tsc --noEmit && vite build`) → clean
- web unit tests (`vitest run`) → **373 passed (43 files)**
- `web/tests/e2e/emergency.spec.ts` in full, on my own port (8470) → **5 passed** (was 4; the new medicine-data test is the 5th)
- `make checkpoint N=14` against my own dev server (port 8470), re-run after every backend round → **passed**, all three times
- No column/schema change, so `tests/test_data_map.py` doesn't apply.

## Reviewer checklists

Both `.claude/agents/*-reviewer.md` subagents were dispatched in the background against the
full diff. This PR does not wait idle on them; one reported before I opened this PR (pasted
verbatim below, with the fix it prompted already applied and re-verified); the other
(`clinical-safety-reviewer`) is applied by hand below since it had not reported yet.

### `plain-words-reviewer` — verbatim report

> I read the full new dict entries in `backend/app/channels/safety_strings.py` (lines 181–190,
> plus the docstring at 192–196) and `docs/plain-words.md` in full. Note: the doc's "## 1. The
> rules" section actually numbers **13** rules, not twelve (1 through 13, ending with "The same
> words every time") — worth fixing in whatever text refers to "the twelve rules."
>
> I also confirmed nothing else in the `EMERGENCY_CARD` dict changed — `ec.medicine`,
> `ec.medicine_when`, `ec.high_risk`, `ec.no_medicine`, `ec.allergy`, `ec.no_allergy`,
> `ec.blood_type`, `ec.chief_who`, `ec.chief`, `ec.no_chief`, `ec.doctor`, `ec.clinic`,
> `ec.insurer`, `ec.ambulance`, `ec.last_reading`, `ec.boundary` are all byte-for-byte what they
> were before; the two new keys are a pure addition at the end of the dict.
>
> **Verdict summary**
>
> | String | en | ms | zh |
> |---|---|---|---|
> | `ec.render_issue` | **FAIL** | **FAIL** | **FAIL** |
> | `ec.render_issue_family` | PASS | PASS | PASS |
>
> **`ec.render_issue` — FAIL in all three languages**
>
> Rules 2 ("one idea"), 3 (word count), 5/6/10/12 (n/a here), and 7 (deferred to the paired
> `_family` line) all pass. The failure is against the governing test in §3 of the doc —
> *"Would his daughter say it to him this way, out loud, as a full sentence?"* — and, by
> extension, Rule 1's spirit ("whole, natural sentences," not stiff ones):
>
> - **en**: `"One line on {name}'s card could not be shown."` — "could not be shown" is a
>   modal-passive rendering-failure phrase (software's point of view: *failed to display*), not
>   something a daughter says about a piece of paper or a spoken card. It reads like an error
>   toast, not family speech.
> - **ms**: `"Satu baris pada kad {name} tidak dapat ditunjukkan."` — same problem: `ditunjukkan`
>   is the formal/bureaucratic passive of *tunjuk*, the register of an official notice
>   ("maklumat tidak dapat ditunjukkan"), not conversational Malay.
> - **zh**: `"{name}的卡上有一行无法显示。"` — `显示` ("display") is squarely a screen/device word.
>   It's the least natural of the three: it wouldn't survive the doc's own Hokkien-aloud check,
>   and no other string in this dict uses tech-register vocabulary for a patient-facing line.
>
> By contrast, the existing (unmodified, presumably-accepted) `ec.no_chief` — `"No family number
> is written down yet."` / `"Belum ada nombor keluarga ditulis."` — shows the house style
> handles a similar "something's missing" idea with concrete, physical verbs ("written down"),
> not abstract UI verbs ("shown/displayed"). `ec.render_issue` should match that register:
> describe the missing content as *left out*, not as a rendering event.
>
> **Rewrite (all three, same "left out" framing so translations stay in lockstep):**
> - en: `"One line on {name}'s card was left out."`
> - ms: `"Satu baris pada kad {name} tertinggal."`
> - zh: `"{name}的卡上少了一行。"`
>
> (`tertinggal` and `少了一行` are both everyday, concrete words — "left behind" / "short by one
> line" — with no bureaucratic or screen connotation, and read naturally spoken aloud.)
>
> **`ec.render_issue_family` — PASS in all three languages**
>
> - **en**: `"Tell your family or the doctor."` — full imperative sentence, one idea, under-ten
>   words (6), no red words, no word to decode. "the doctor" used generically has direct
>   precedent in the unmodified `ec.boundary` ("not a doctor's advice"), so it's consistent with
>   existing house style rather than a Rule 13 violation.
> - **ms**: `"Beritahu keluarga anda atau doktor."` — direct imperative, plain vocabulary,
>   matches the en structure.
> - **zh**: `"请告诉家人或医生。"` — I initially flagged the polite `请` as too formal for family
>   speech, but the unmodified `ec.chief` entry (`"Call {chief} first."` → `"请先打给{chief}。"`)
>   already establishes `请` as this dict's standing convention for spoken Chinese imperatives on
>   this card. So this is consistent, not a deviation — pass.
>
> **One more observation:** the docstring at line 192–196 says these two lines "cannot
> themselves fail the standard and go silent a second time" because they're "static, no clinical
> content." That's true of clinical-content risk, but a string can still fail the plain-words
> *register* test (passive/tech phrasing) without containing any clinical content — so I'd treat
> "static" as necessary but not sufficient for passing plain-words, and would still want a
> native-speaker pass on the ms/zh rewrites before ship, per the doc's own caveat that its
> Malay/Chinese are "a first translation awaiting a native speaker's pass."

### Fixed during review

Took the reviewer's rewrite verbatim (`ec.render_issue` only — `ec.render_issue_family` passed
as written):

- en: `"One line on {name}'s card was left out."`
- ms: `"Satu baris pada kad {name} tertinggal."`
- zh: `"{name}的卡上少了一行。"`

Re-verified with `app.safety.plain_words.verify()` directly (0 findings), `make plain-words`
(0 failures), `make language` (0 failures), and re-ran the targeted pytest set (192 passed) — all
clean with the new wording, all numbers above already reflect it. The reviewer's aside about the
native-speaker caveat on ms/zh stands as a general property of this whole file (unrelated to
this PR — every line in `docs/plain-words.md` carries that caveat already) and their "twelve vs
thirteen rules" note is about the doc's own heading, not anything in this diff — flagging both
here rather than fixing either, since neither is this PR's file to fix.

### `clinical-safety-reviewer` — applied by hand (had not reported when this PR opened)

1. Every profile read goes through the keys module with a key context. **PASS** — no new reads
   anywhere in the diff; `_medicine_label`/`_plain_medicine` only reshape data `_projection()`
   (unchanged) already fetched under `KeyContext`/`Scope.EMERGENCY`.
2. Every new Fact carries provenance; every Card records its State. **PASS / N/A** — no new
   Facts; `EmergencyCard.state_id` still comes from the unchanged `render_from_state` /
   `render_from_last_snapshot` call. The two new lines are composed into the same row's
   `lines`, not separate persisted rows.
3. No output starts, stops or changes a medicine; treatment-changing content is rerouted as a
   doctor question. **PASS** — the new lines are purely informational ("One line ... was left
   out.", "Tell your family or the doctor."); nothing in the diff instructs starting, stopping
   or changing a medicine.
4. Drug identification, interactions and dosing come from the licensed data client only.
   **PASS** — `_plain_medicine` still calls `registry.monograph(line.generic)` through the
   `DrugRegistry` port; `_medicine_label` only decides whether to *display* the already-licensed
   generic name beside the plain one, never invents pharmacology.
5. Red-flag paths escalate before any ranking or caps. **N/A** — diff doesn't touch
   red-flag/delivery code.
6. High-risk drugs require a label photo before a dose is saved. **N/A / PASS** — diff doesn't
   touch the write path (`app/safety/high_risk.py`'s hook is untouched); the `ec.high_risk`
   marker line is preserved and, per the new test, still renders for a no-story high-risk drug.
7. Health data stays in region; no health data reaches analytics. **PASS** — no new I/O.
8. Consent is recorded at claim and per key; refusals are visible in the audit. **PASS / N/A** —
   unaffected.

Required changes: **none**. Two non-blocking notes for the record: (a) `_medicine_label` does a
`render()` dry run and, on success, the real `say()` call renders the same text again — cheap
and pure, flagged only for awareness; (b) the no-story fallback relies on `plain_words.py`'s
existing rule that a capitalized non-initial word reads as a name — the same mechanism already
used for doctor/chief/insurer names, not a new loophole, and it doesn't touch `GLOSSARY`.

I will paste the actual `clinical-safety-reviewer` agent's report as a PR comment if/when it
reports, since the by-hand pass above stands in for it here rather than blocking the PR on it.

- Did not add any of the 46 new generics to `plain_words.py`'s `GLOSSARY` — the issue explicitly
  rules this out as a workaround, and it wouldn't scale to the next 50.
- Did not weaken `_check_length`/rule 3 itself, and did not add `"medicine"` to `NAME_SLOTS`
  (which would have silently exempted every other template that fills a `medicine=` slot —
  `nfw.not_taken`, `nfw.ask_before`, `notice.not_taken` — from content verification, well beyond
  the emergency card). The new `_medicine_label` fallback is scoped to exactly the two emergency
  card templates that need it.
- Did not touch `docs/parity.md`'s E13-01 gap line (no insurer-field / one-card-per-language /
  no-web-screen gaps) — those are separate, unrelated open items; this PR is a safety bug fix,
  not that story's completion.
- Did not add a `docs/adr/` entry — the design decision (data-not-sentence for the register's
  name; opportunistic parenthetical; static safety-net lines) is written down in the code itself
  (`_plain_medicine`, `_medicine_label`, `Medicine`, `compose_lines`, and the new
  `safety_strings.py` docstrings), matching this file's own convention of narrative doc-comments
  rather than a separate ADR for a targeted fix.

## Round 2: fixing the clinical-safety review's three required changes, and the independent reviewer's finding

`.claude/agents/clinical-safety-reviewer.md`, run against the PR as opened, listed three
required changes. All three are done, re-verified, and pushed:

1. **The probe checks a different string than the one that ships.** `_medicine_label` used to
   probe `ec.medicine` with the pre-`theirs()` string, while `compose_lines` applied `theirs()`
   afterwards on the chosen label. Fixed: `theirs()` now runs inside `_medicine_label`, before
   the trial render, so the probe checks exactly the string that ships (English "your" becomes
   "{name}'s", which changes both the words and the length).
2. **The probe only checked `ec.medicine`.** `ec.high_risk` fills the same label in a longer
   sentence ("{name}'s doctor watches {medicine} closely.") and sits closer to rule 3's 15-word
   hard fail; before this, they avoided collision only by coincidence, because `GLOSSARY`'s
   enrichable set doesn't overlap `HIGH_RISK_CLASSES`. Fixed: `_medicine_label` now probes every
   template in the new `_MEDICINE_TEMPLATES` tuple (`ec.medicine`, `ec.high_risk`), and falls
   back to the plain label if *any* of them would fail.
3. **Test both together.** New test `test_ec_high_risk_is_probed_too_not_only_ec_medicine`:
   a synthetic high-risk generic with a story, whose enriched label passes `ec.medicine`'s probe
   (12 words, under the "Ash"-checked threshold) but fails `ec.high_risk`'s (16 words, over it)
   — combined with a real, long, multi-word patient name ("Muhammad Firdaus Abdullah") sized so
   the *fallback* label still renders cleanly in both templates with the real name substituted,
   not just the short `NAME_STAND_IN` the probe itself checks against. Asserts `ec.high_risk`
   renders, is never dropped, and the general safety net (`ec.render_issue`) never had to fire.

An independent reviewer then found the bigger gap answered above (the live screen never showed
the chemical name at all) and one more asymmetry:

4. **`ec.render_issue`/`_family` bypassed `say()`'s try/except**, calling `render()` directly —
   a future edit to either template would have raised out of card generation entirely instead of
   degrading. Fixed: both now go through `say()` like every other line, so they fail the same
   way (dropped, logged) rather than crashing card generation if they're ever wrong.

New web test, `web/tests/e2e/emergency.spec.ts`'s `"the medicine data: ..."`, walks the
**rendered screen**, not the JSON — the independent reviewer's own point, since a JSON test is
how the chemical-name gap got through the first time. It seeds frusemide (glossary-safe),
paracetamol (a real #202-era generic whose sentence can't carry its name), and warfarin
(high-risk), and asserts all three chemical names and strengths are visible on the rendered
card, that the sentence for paracetamol specifically never says "paracetamol" (proving the data
row, not the sentence, carries it), and that the high-risk marker shows for warfarin. It then
seeds two more owners — one Malay-language, one Chinese-language — and asserts the same data row
renders identically for each. `seedOwner`'s test helper gained an optional `language` parameter
(default English, so every other existing caller is unaffected) to make this possible.

**A note on the "no plain-name story" scenario specifically:** the fixture registry
(`tests/fixtures/drugs/registry.json`) happens to have a story for all 50 of its generics today
(established in the original measurement above), so an e2e test genuinely cannot construct that
exact case through the real ingestion API — a name the registry cannot identify is refused before
a `MedicationLine` is ever written. That case is covered at the unit level instead
(`test_a_high_risk_medicine_with_no_story_still_appears_with_its_marker`,
`test_ec_high_risk_is_probed_too_not_only_ec_medicine`), which is where it's actually reachable.

## Risk

- The printable page's medicines table now shows a `(generic)` caption for every medicine that
  has a plain-name story (previously it echoed whatever was baked into `plain_name`, which for
  4 glossary drugs already included it and for the other 46 never did) — a visual change to a
  printed/PDF-like artefact. I did not find an existing snapshot test pinning the printable
  page's exact HTML; `test_the_printable_page_and_the_live_card_show_the_same_medicines` covers
  content, not exact markup.
- I relied on the fixture register (`tests/fixtures/drugs/registry.json`) for the registry-wide
  test; a real licensed registry adapter (when one exists, per `DrugRegistry`'s Phase-0-stub
  note) could have different plain-name coverage, but the fix's guarantee (medicine name never
  fails verification) is independent of which generics have stories, so it should hold either
  way — worth a second look when a real adapter lands.
- The high-risk marker in the new data row (" · high-risk") is literal English in every
  language, deliberately not run through plain-words since it's data, not a sentence to him —
  but a Malay- or Chinese-reading family member glancing at his own card will see an English
  word there. Worth a follow-up if that reads as inconsistent in practice; I left it in English
  rather than invent a translated data-label with no existing convention to match.
- Per the plain-words reviewer's caveat: the ms/zh wording throughout `docs/plain-words.md`,
  including my new lines, is "a first translation awaiting a native speaker's pass" — true of
  this whole file, not introduced by this PR, but worth a human check before this ships wide.


ashfrnndz21:
Addressed the three defects from the clinical-safety review (its post to the PR failed, but the findings were real) in 755c2366.

**A1 — `backend/app/safety/emergency_card.py:423-425`.** `_medicine_label`'s fallback branch (`language != "en"`, or `has_plain_name` false) was handed straight back to the caller with no probe, unlike the enriched label. A generic that is also one of `GLOSSARY`'s red words needing its plain name beside it (frusemide, amlodipine, atorvastatin, metformin) fails rule 4 on its own, in any language, the moment the register has no story to pair it with — so the sentence naming it could be silently withheld. Fixed by probing the fallback the same way and falling back further, to the existing `YOUR_MEDICINE` catalogue phrase, before anything reaches `say`. Test: `test_the_fallback_label_is_probed_too_not_only_the_enriched_one` — a medicine with no plain name, in en/ms/zh, still appears on the card via `ec.medicine` and `ec.high_risk`, never `ec.render_issue`. Commit 755c2366.

**A2 — `backend/app/channels/printable.py:105-109`.** The printable page paired a medicine's English twin by its position among `ec.medicine` lines, not by which medicine it names — the patient-language and English lines are two separate passes over the same medicines with nothing guaranteeing the same length or order (a withheld line per A1 is one way they diverge, not the only one). Fixed by carrying the medicine's own `line_id` on `Line` (`medicine_line_id`, set by `say`/`compose_lines`) and joining by it in printable's `twin()`. Test: `test_the_printable_page_pairs_the_english_twin_by_medicine_not_position` — two medicines, only the second survives in his own language, both survive in English; the twin must be the second medicine's own English sentence, never the first's. Commit 755c2366.

**A3 — `web/src/screens/Emergency.tsx:55`.** The literal `" · high-risk"` was patient-visible English emitted outside the verified `render()` path. Fixed by moving the word into the backend: a new catalogued, plain-words-checked phrase `HIGH_RISK_LABEL_WORDS` (en/ms/zh) in `safety_strings.py`, surfaced as a per-medicine `Medicine.high_risk_label` / `MedicineOut.high_risk_label`, and the screen now renders only `medicine.high_risk_label` (only the separating " · " is the client's). The null-`strength` handling on the data row is unchanged. Commit 755c2366.

Gates run: `pytest backend/tests/test_emergency_card.py backend/tests/test_emergency_card_insurer.py backend/tests/test_safety_api.py backend/tests/test_demo_mode.py backend/tests/test_consent_withdraw_api.py` (65 passed); `make lint` (clean); `make plain-words` (0 failures); `make language` (0 failures); `cd web && npx tsc --noEmit` (clean); `npx vitest run tests/unit/emergencyCache.test.ts` (8 passed, the one existing unit test touching the changed types).

PR #224, opened 2026-09-17
Everything that can reach a person goes through the one review queue (#181, #188, #145)

ashfrnndz21:
Closes #181, #188, #145.

## Update: a blocking finding from independent review, fixed

An independent review of this PR found that #181's own fix had a hole: in
`backend/app/delivery/feed/search.py`'s `SAFETY` branch, a notice whose words would start, stop
or change a medicine wrote a `CardType.QUESTION`/`DeliverTo.MEMO` `FeedItem` and then
**`continue`d past the code that writes the caregiver's notice** — so the caregiver was never
told, whether or not the batch matched his own box. And nothing reads `DeliverTo.MEMO`:
`rank.py`'s two supplies exclude it, no route queries it, and the real doctor's memo is a
separate table `search.py` never wrote to. The most dangerous case — a confirmed recall on his
own box that also reads as a treatment change — reached nobody: not him (correctly), not his
chief, not his doctor.

**Fixed:** the caregiver notice is now unconditional — the treatment-change reroute is additive,
never a substitute for it. A treatment-changing notice is *additionally* filed as a real doctor
question through `reasoning.visits.memos.write_memo` (the same door the post-visit summary
uses, read back by `current_memos` — what `brief.py` and `questions.py` themselves call), never
a `FeedItem`. New `MemoSource.SEARCH` and an `ask_safety_notice` catalogue key (en/ms/zh). A
secondary-write failure here is caught in a savepoint of its own (mirroring `items._sample`) so
it can never cost the caregiver notice. The test was rewritten to prove delivery, not row
existence: a second, independently resolved `KeyContext` for a caregiver reads the notice back
through the real `feed_page`, and `current_memos` reads the question back through the real
function his doctor's brief calls — covering the confirmed-his-own-box case explicitly, which is
the one that was broken.

Filed two follow-ups rather than fixing them here (both out of this PR's scope): **#230** — the
review queue's sample budget is one global counter with no profile and no time window, which
#188 can exhaust deployment-wide; **#231** — the *general* (non-safety) branch has the identical
`FeedItem`/`DeliverTo.MEMO` dead end this fix closed for the safety-notice branch only. A second
independent-review pass, run specifically to verify this fix, confirmed the fix itself (7 of 8
checks pass, the required change satisfied) and independently found the same #231 gap, scoring
it "required, same severity" on its own judgment — see "Open finding" further down for why it is
still on #231 rather than fixed in this push, and that a Malay wording issue in the new
[...]

Issue #10, opened 2026-09-13
E01-02 Health biography session: guided capture of existing papers into the record with


ashfrnndz21:
**Epic** E01 Onboarding and health biography
**Persona** Chief · **Tier** T1 · **Priority** P1 · **Size** L

**Story**
Health biography session: guided capture of existing papers into the record with agent assistance

**Acceptance**
A shoebox of 30 documents becomes a reconciled record in one sitting

**Depends on**
E02-01,E02-02

**Spec** see docs/00-MASTER-BUILD-SPEC.md section 0 for the document that covers this epic.

PR #236, opened 2026-09-17
A treatment-changing finding never reaches nobody (#231)

### LINKED ISSUE #231: A found page's treatment-change reroute (the general branch) also reaches nobody
Found while fixing #181's blocking review (PR #224) — the same defect the review caught in the
`JobKind.SAFETY` branch, still present in the general (non-safety) branch a few lines below it.

**The bug.** `backend/app/delivery/feed/search.py`'s general found-page path — every job kind
except `SAFETY` (explainer, local, seasonal, food, clip, worth-knowing) — reroutes a page whose
words would start, stop or change a medicine to a `CardType.QUESTION` `FeedItem` with
`deliver_to=DeliverTo.MEMO`:

```python
if changes_treatment([compressed.headline, *compressed.body]):
    question = await create_item(..., type=CardType.QUESTION, deliver_to=DeliverTo.MEMO, ...)
```

Nothing reads it, the same way #181's safety-notice version didn't:

- `rank.py`'s `_patient_supply` excludes anything not `DeliverTo.PATIENT`.
- `rank.py`'s `_caregiver_supply` includes only `PATIENT` and `CAREGIVER`, never `MEMO`.
- No API route queries `FeedItem` for `deliver_to == MEMO` or `type == CardType.QUESTION`.
- The real doctor's memo is a separate table (`app.reasoning.visits.memos.Memo`, written only
  from `visits/summary.py`); `search.py` never calls `write_memo`.

So a gap or an evergreen/local/seasonal/food page whose words happen to read as a reason to
start, stop or change a medicine is silently dropped — a row is written, a row nothing reads.

**Fix applied to the SAFETY branch only, in #224:** replaced the `FeedItem`/`DeliverTo.MEMO`
write with a real `reasoning.visits.memos.write_memo` call (`kind=MemoKind.ASK`), which *is*
read by `current_memos` (what `brief.py` and `questions.py` call) and by
`questions.questions_for` (what the doctor's brief actually shows). The general branch's
`changes_treatment` reroute (`search.py`, a few lines below the `SAFETY` block) was left as it
was — same shape, same bug, but out of scope for #181/#188/#145, so filed here rather than
changed inside that PR.

**Suggested fix:** the same one — route the general branch's reroute through `write_memo` too,
with an appropriate `MemoSource` and catalogue key per job kind (or a shared one, if the wording
can stay accurate across kinds), instead of a `FeedItem` nobody reads.

Not fixed in #224 — flagged per the operator's instruction not to expand that PR's scope.

ashfrnndz21:
Closes #231.

Follow-up to #224. The general `changes_treatment` branch in `search.py` wrote a `QUESTION`/`MEMO` feed item that nothing reads, so a treatment-changing finding from a food, seasonal or local job reached no one.

- Routed by job kind: a medicine he takes becomes a real doctor question through `write_memo`; food, season or local hazard is held for his chief.
- A safety notice whose wording fails plain words still reaches his chief as a fixed line, instead of being dropped silently.
- Tests check delivery, not that a row exists.

Built by a builder whose own reviewer passes were still finishing when it stopped; opened by the operator from its pushed branch. The `review` check is out of credits and not a gate.


ashfrnndz21:
Addressed the review's findings in bc99f5e.

- **FAIL A** (`search.py:630`) — `items.create_item` now refuses any card whose words would start, stop or change a medicine, for every `DeliverTo`, before anything else (`TreatmentChangingCard`). `search.run_job` no longer builds the caregiver's card from `compressed`/notice words when a finding is treatment-changing (SAFETY branch and the general learning/local/seasonal/food branch both fixed) — it uses the new `needs_doctor_look_lines` instead, which names that a question was filed for the doctor and never repeats the finding.
- **FAIL B** (`search.py:655-692`) — the dead `NotPlainWords` fallback branches are removed: `create_item` only verifies plain words for `DeliverTo.PATIENT`, and treatment-changing content is now never routed to `create_item` in the first place (it goes straight to `needs_doctor_look_lines`), so the branch that caught a `NotPlainWords` raised on a caregiver card was provably unreachable. A doctor question is now filed whenever a finding is treatment-changing, not only when a real drug name is known — `_ask_the_doctor` files a name-free question (`ask_medicines_change`, the same key `visits.summary.compose_items` already uses) instead of nothing.
- **FAIL C** (`review.py:381`) — `sample_card` now samples a caregiver-delivered card of a `KEPT_AS_WRITTEN` type (learning, clip, local, seasonal, food, notice), not only a safety notice, so a rerouted treatment-changing card reaches the pharmacist's first-fifty sample.
- **FAIL D** (`search.py:707-720`) — `_key_for` is keyed on the job, not just the page and kind, so two jobs that find the same page can no longer race on whether a question gets filed (each job's dedupe is now independent of every other job's, and still idempotent against its own earlier run). `compose._learning`'s job read is also given a deterministic order (`created_at`, then `id`) as a second guard.
- **Provenance** (`search.py:544-550`, `:669-675`) — the reroute's `Why` now carries real `fact_ids` and a real `plain` sentence in both branches, instead of an empty string.
- **Now uncovered** — added `test_treatment_changing_words_never_reach_any_card_for_any_audience`, which asserts `TreatmentChangingCard` for every `DeliverTo` and that no `LEARNING` card exists afterwards.
- **Plain words** — rewrote the reroute copy (was `"notice_fallback"`: "A notice needs a look" / "Open Nura to see it.") as `needs_doctor_look_lines`: names Nura and the doctor, says a question was already filed, never asks him to open the app he is already in. In en/ms/zh, with `*_THEIRS` twins added. Also reworded `learning_keep_taking` ("Ask {doctor} before you stop this medicine.") — it was tripping the same `changes_treatment` check it exists to guard against, once the check ran over the fully-assembled card rather than only the raw compressed text; it's now "Ask {doctor} before you stop taking it." with the same meaning.

Also fixed while in the area, since they were in the review's own findings list:
- `_medicine_he_takes`'s registry fallback asked the register "is this a generic?" (`LabelFields(generic=written)`), which could only ever repeat the fact the line above it had already checked and returned False on — a no-op. It now asks by brand, so a brand-named watch (e.g. "Marevan" for warfarin) on a medicine he takes resolves.
- `_ask_the_doctor`'s own audit write (logging a filing failure) is now guarded in its own `try/except`, so a second, unrelated failure writing that down can no longer escape and cost him the caregiver notice.

Two things surfaced only once this was pushed, flagged by the operator from CI on this same head:
- `tests/test_feed_api.py::test_sources_and_search_jobs_are_the_owners_and_allowlist_scoped` was asserting the pre-#224 shape (a `CardType.QUESTION`/`DeliverTo.MEMO` `FeedItem`, and only 1 item made from a job that finds 2 usable pages) — stale since #224 removed that `FeedItem` shape entirely. Rewrote it to assert the current one: 2 items (the food card for him, the rerouted card for her), a real `Memo` (name-free, since no warfarin fact is on his record in that test), and the rerouted card pointing at it.
- `web/tests/e2e/day.spec.ts`'s red-flag-ordering spec was reported failing on this head; ran it locally (`day.spec.ts`, all 15 tests, chromium) against this branch and it's green — nothing in this fix touches that flow, and I couldn't reproduce the failure.

Gates run and green: `pytest backend/tests/test_feed.py backend/tests/test_feed_api.py backend/tests/test_feed_formats.py backend/tests/test_review_queue.py backend/tests/test_data_map.py` (116 passed), `make lint` (ruff + mypy, 0 issues), `make plain-words` (0 failures), `make language` (0 failures), `web/tests/e2e/day.spec.ts` (15 passed).

Not addressed from the review's "Required changes" list: #10 (reconciling `docs/health-feed-spec.md`'s stale prose and the `items.py:100-103` comment about the now-dead `CardType.QUESTION`/`DeliverTo.MEMO` shape) — I fixed `search.py`'s own module docstring since it was the most directly misleading, but left the doc and the other comment for a follow-up rather than widen this diff further.

Issue #13, opened 2026-09-13
E02-01 Photo to OCR to structured extraction for labs, reports, prescriptions and recei


ashfrnndz21:
**Epic** E02 Capture and ingestion
**Persona** Chief · **Tier** T1 · **Priority** P1 · **Size** L

**Story**
Photo to OCR to structured extraction for labs, reports, prescriptions and receipts

**Acceptance**
Lab values extracted with unit and date at 95% on a labelled MY/SG test set

**Depends on**
E00-03

**Spec** see docs/00-MASTER-BUILD-SPEC.md section 0 for the document that covers this epic.

PR #238, opened 2026-09-17
Claude-backed document extractor, demo-only (E02-01 gap)

ashfrnndz21:
## What this closes

`docs/parity.md`'s gap line for **E02-01**:

> fixture extractor only, 9 placeholder papers (2 lab reports), read accuracy asserted < 1.0
> and never against 95%.

This PR does not attempt the 95%-on-a-labelled-set acceptance line (that needs a real labelled
MY/SG set and live calls, both out of scope for a no-live-call PR — see "What I did not
build"). It replaces the other half of the gap: **today a real PDF or photo comes back from
`FixtureExtractor` almost empty**, because it only recognises the handful of redacted sample
papers it ships with. This PR adds a second, real adapter behind the same port
(`app.ingestion.extract.Extractor`) so an uploaded document is actually read.

No related open GitHub issue names this specific gap (#13/E02-01 is closed on the fixture
stub already); this PR was scoped directly from the operator's brief.

## What I built

- **`backend/app/ingestion/claude_extract.py`** — `ClaudeExtractor`: reads a PDF or a JPEG/PNG
  photo with `client.messages.create(model="claude-opus-5", ...)`, a `document` or `image`
  content block placed **before** the text block, structured output via
  `output_config={"format": {"type": "json_schema", "schema": ...}}` (never the deprecated
  `output_format`), and `stop_reason` checked before any content is read — `"refusal"` and an
  answer that doesn't parse to the schema both read as `Extraction.nothing()`, the same honest
  "could not read this page" answer `FixtureExtractor` gives for bytes it doesn't know. Maps
  onto the port's exact field shape (subject, attribute, value, unit, confidence, unreadable)
  with a per-field confidence; a field whose confidence the model can't be trusted on (missing,
  wrong type, out of 0-1) is kept at confidence 0.0 - held for his confirmation, never
  invented. No `budget_tokens`, no assistant prefill. Nothing is logged but that a call
  happened and how it ended - never document bytes, the prompt, or the model's answer.
- **`backend/app/ingestion/extract_provider.py`** — `extractor_for(settings)`: `fixture` (the
  default, unchanged) or `claude`, chosen by `NURA_EXTRACTOR`. `claude` refuses to build unless
  `NURA_DEMO_MODE=1` (ADR 0008) - Anthropic's first-party API doesn't process in SG or MY, so
  this adapter may only run where every document is test data by declaration, and a plain dev
  run is deliberately **not** enough (`test_claude_refuses_to_build_on_a_plain_dev_run_too`). It
  also refuses to build without a key (`NURA_ANTHROPIC_API_KEY` or the SDK's own
  `ANTHROPIC_API_KEY`) - both checked with a clear `RuntimeError`/`MissingSetting`, not a bare
  network failure later.
- **`backend/app/settings.py`** — `NURA_EXTRACTOR` (default `fixture`) and
  `NURA_ANTHROPIC_API_KEY`, read the same way every other provider setting is.
- **`backend/app/main.py`** — `providers_for` now calls `extractor_for(settings)` instead of
  hard-coding `FixtureExtractor`; nothing else in the wiring changes.
- **`backend/pyproject.toml`** — pins `anthropic==1.6.0`.
- No changes to `review.py`, the review card, `CONFIDENCE_THRESHOLD` (0.8), the drug registry,
  or the high-risk-label rule - every caller of the port is untouched, as the modularity rule
  in `CLAUDE.md` requires. Checkpoint 5 (below) proves the default path is bit-for-bit the same.

## To turn it on

```
NURA_DEMO_MODE=1
NURA_DEMO_LOGIN_CODE=<six digits>
NURA_EXTRACTOR=claude
NURA_ANTHROPIC_API_KEY=<the key>   # or plain ANTHROPIC_API_KEY
```

Anywhere `NURA_DEMO_MODE` is not `1` - including a laptop dev run - `NURA_EXTRACTOR=claude`
refuses to start (`ClaudeExtractorOutsideDemo`), so it can only ever run against demo data:
test phone numbers, wiped nightly, banner on every screen (ADR 0008's existing promises).

**Cost note.** Every upload through `POST /profiles/{id}/imports` and the photo capture route
bills one document to the key's Anthropic account. There is no cap or budget in this PR - a lot
of uploads to a demo is a lot of calls. Worth a rate limit or a daily cap before this runs
unattended on a public demo link.

## What I did not build

- **The 95% accuracy line itself.** That needs a labelled MY/SG test set and live calls against
  it; this PR is entirely mocked, per the "no live calls" gate. `tests/paper_accuracy.py` still
  runs only against the fixture.
- **An in-region provider.** Still nobody's built one; the seam (`extractor_for`) is exactly
  where a third adapter goes, and nothing above it would need to change.
- **Per-field bounding boxes / page localisation.** Claude isn't asked for a box, and I didn't
  fabricate one (`span=None` throughout) - inventing a plausible-looking box the model didn't
  actually give felt worse than leaving it out. The review card doesn't require a span; only
  the multi-page "which page" annotation is lost.
- **HEIC/WebP photos.** The brief names JPEG/PNG only for the image block; `PHOTO_CONTENT_TYPES`
  allows HEIC/WebP too, and those now hit the same honest `Extraction.nothing()` an unknown
  fixture digest gives, rather than being converted or rejected outright.
- **A call budget/rate limit.** See the cost note above.
- **`docs/parity.md` / `docs/checkpoints.md` updates.** Left alone since the fixture stays the
  default and checkpoint 5's fixture-mode results are unchanged; happy to update the parity
  row's status in a follow-up once the accuracy work above lands.

## Gates (all clean; only the extractor/capture/import test files, not the full suite)

Ran: `tests/test_claude_extractor.py` `tests/test_ingestion.py` `tests/test_ingestion_api.py`
`tests/test_imports.py` `tests/test_handwriting.py`, plus `tests/test_fixture_refusals.py` and
`tests/test_demo_mode.py` (the fixture/demo-mode residency gate I added into), and
`tests/test_data_map.py` (no new stored column, so N/A, but ran it anyway) - **97 passed**.

```
make lint          -> ruff: all checks passed; mypy: no issues found in 266 source files
make plain-words    -> 8031 strings in 33 files, 0 failures, 120 notes (all pre-existing, none mine)
make language       -> 8170 strings, 2752 keys in 31 catalogues, 0 failures, 89 notes (all pre-existing, none mine)
```

`make checkpoint N=5` (own port 8510, throwaway sqlite): **checkpoint 5 passed** - identical
output to before this change (default `NURA_EXTRACTOR=fixture` path untouched).

Full backend suite left to CI's `backend`/`backend-postgres` jobs, per the shared-machine rule.

## Reviewer reports

Run by hand over the diff, following `.claude/agents/clinical-safety-reviewer.md` and
`.claude/agents/plain-words-reviewer.md`'s own briefs - no custom subagent of that name is
available in the harness this session runs under, so I read each agent's brief and performed
its checklist myself against the actual diff.

### clinical-safety-reviewer

1. Every profile read goes through the keys module with a key context. **Pass** - unrelated;
   `ClaudeExtractor.extract` takes only `(data, content_type, hints)`, no DB access; every
   route calling it is unchanged.
2. Every new Fact carries provenance; every Card records its State. **Pass** - unrelated;
   `review.py`, which writes Facts and Cards, is untouched.
3. No output starts, stops or changes a medicine; treatment-changing content is rerouted as a
   doctor question. **Pass** - the system prompt (`claude_extract.py`, `_SYSTEM_PROMPT`)
   explicitly forbids advice/diagnosis/treatment instructions, and structurally the model can
   only emit the fixed schema (subject/attribute/value/unit/confidence/unreadable) - there is
   no free-text field a treatment instruction could travel in.
4. Drug identification, interactions and dosing come from the licensed data client only.
   **Pass** - `ClaudeExtractor` proposes a `medicine`/`name` text field read off the label the
   same way the fixture does; identification against the registry still happens downstream in
   `app.medicines.service`, untouched.
5. Red-flag paths escalate before any ranking or caps. **N/A** - not touched.
6. High-risk drugs require a label photo before a dose is saved. **Pass** - unaffected;
   `app/safety/high_risk.py`'s hook checks the artefact's kind (PHOTO), not which extractor
   produced the card; checkpoint 5 re-confirmed the warfarin-label path unchanged.
7. Health data stays in region; no health data reaches analytics. **Conditional pass - the main
   risk in this PR.** `ClaudeExtractor` sends document bytes to Anthropic's first-party API,
   which is not in SG or MY. This is only safe because `extractor_for` refuses to build it
   outside `NURA_DEMO_MODE=1` (tested, including the "a dev run alone is not enough" case) -
   i.e. only test data, by ADR 0008's own declaration, ever reaches it. **This adapter must
   never be enabled on a deployment holding real family data.** Flagging this explicitly rather
   than calling it a clean pass.
8. Consent is recorded at claim and per key; refusals are visible in the audit. **Pass** -
   `require_consent` on the upload routes is untouched.

Required changes: none, on condition that `NURA_EXTRACTOR=claude` is never set outside
`NURA_DEMO_MODE=1` - enforced in code and tested, not just documented.

### plain-words-reviewer

No patient-facing strings were added or changed. The system/user prompts in
`claude_extract.py` are sent to the model, not shown to Pa or Mei; the review card, its
copy, and every string in `web/` and `app/safety/` are untouched. **N/A - nothing to review.**

## Risks worth the operator's attention

- The residency gate (`NURA_DEMO_MODE=1`) is the only thing standing between this adapter and
  sending real health data to a US/global API. It's enforced in code and covered by tests, but
  it is a single boolean - worth a second look before anyone flips `NURA_EXTRACTOR=claude` on
  a non-demo deployment.
- No spend cap: a demo link left open bills per upload with no ceiling.


ashfrnndz21:
Addressed every FAIL from the last review (2026-09-17T05:47Z), across three commits now on this branch:

- **CI red (`anthropic==1.6.0` missing from requirements.lock)** — `8ec8d822` pins it plus its missing transitive deps (distro, jiter, docstring-parser, cryptography, etc). Verified in a scratch venv: `pip install --no-deps -r requirements.lock && pip check` → "No broken requirements found." `tests/test_requirements_lock.py` passes.
- **Region (must fix)** — `8fbef54c` adds `docs/adr/0017-claude-runtime-features-demo-only.md`: Claude-backed runtime features (extractor now, searcher/compressor later) run only under `NURA_DEMO_MODE=1` on demo/test data; real patient data never leaves the region. `claude_extract.py` and `extract_provider.py` docstrings now cite ADR 0017, not ADR 0008, as residency authority.
- **Consent and audit** — `3b2a89e3`: `review_artifact` now writes a distinct `SHARE`/`EXTERNAL_MODEL_PROCESSOR` audit line whenever `extractor.external_processor` is set, before the extraction call, so the trail tells an external-model read apart from a fixture read (`app/ingestion/review.py`). `docs/trust/recording-consent.md` names the demo-only processor (Anthropic) beside the recording pattern's own third parties.
- **Provenance (`page` always null)** — `3b2a89e3`: the schema now asks for `page` per field; `_span_of` turns a trustworthy page into a whole-page `Span`, carried through to `ExtractedField.span`. Untrustworthy pages still reach the card (no locator), not dropped.
- **A misread page looks empty** — `3b2a89e3`: `_IMAGE_TYPES` now matches the route (`app/ingestion/photos.py`) minus HEIC — jpeg/png/webp accepted; HEIC is refused with a log line and a new `Extraction.unsupported_file_type()` outcome, distinct from `nothing()`, surfaced as its own honest notice (`photo_kind_not_read`, en/ms/zh) instead of the untrue "could not read this page, try again in daylight" line. `MAX_TOKENS` raised 4096→8192; `stop_reason == "max_tokens"` is its own logged case, not a fallthrough to parse failure.
- **`backend/CLAUDE.md:8`** — `3b2a89e3` adds `app/llm/`: `client_for` (one client factory, now with a 60s timeout) and `load_prompt`; the extraction prompt moved to `app/llm/prompts/extract_document.txt`. `extract_provider.py` and `claude_extract.py` both call through it.
- **Lint** — the over-100-char lines are gone; `make lint` (ruff + mypy) is clean.
- **Plain words** — the untrue string is fixed via the existing `Notice` mechanism rather than a new web-strings key: `DocumentKind.UNSUPPORTED_FILE_TYPE` → `Notice.PHOTO_KIND_NOT_READ` → `card.notice` (`app/channels/strings.py`, en/ms/zh), which `Records.tsx` already prefers over the `kindUnknown` fallback.

Gates run locally (scratch venv, backend/.venv, gitignored):
- `pytest -q tests/test_claude_extractor.py tests/test_requirements_lock.py tests/test_handwriting.py tests/test_paper_accuracy.py` → 47 passed
- `make lint` → ruff clean, mypy "Success: no issues found in 269 source files"
- `make plain-words` → 8037 strings, **0 failures**, 120 pre-existing notes (unrelated)
- `make language` → 8176 strings, **0 failures**, 89 pre-existing notes (unrelated)
- `pip install --no-deps -r requirements.lock && pip check` → clean

Not run locally per the shared-machine rule: the full `make test` suite — left to CI's `backend`/`backend-postgres` gate.

PR #233, opened 2026-09-17
RE-02 + RE-01: the feeling note reaches the visit; evidence and candidate types

ashfrnndz21:
Design doc: `docs/recommendation-engine.md` and `docs/adr/0016-correlation-sits-beside-state.md` (main `e1f9c28b`). Owner's decisions: PR #223's comment. No GitHub issue exists yet for RE-01/RE-02 specifically (they are the first two stories of the recommendation-engine build plan, §6, Wave 0) — nothing to close.

Built in order: **RE-02 first** (the broken promise), then **RE-01** (evidence and candidate types). One PR: each is small, and RE-01's `private_to`/`Evidence` mechanism is the same "one door" shape RE-02 leans on for `FeelingNote`, so reviewing them together is short and they read as one story about provenance.

## RE-02 — the feeling note reaches the visit

> Nura breaks a promise it makes to the patient. When he logs a feeling note, Nura tells him "Nura will keep this for your visit to Dr Tan." But nothing in visit prep reads that note, and the tap does not write the symptom record the pre-visit brief reads.

**Acceptance (docs/recommendation-engine.md §6, RE-02):** *"a cloud tap read against a new medicine appears on that visit's questions and brief, citing the note"*

Answered:
- `app/reasoning/visits/models.py`: `QuestionSource.FEELING`.
- `app/reasoning/visits/questions.py`: `feeling_notes_for()` reads every `FeelingNote(outcome=FOR_THE_DOCTOR)` tied to *this* appointment's own `appointment_id` (ADR 0004's one-door rule: every scope the note's reasons rest on, via `note_scopes`, not just `Scope.RECORDS`). `question_from_feeling()` turns each into a `Question` row, using the note's own already-verified line as-is — never re-templated, so the words he was shown on the cloud's reply are the words that reach the doctor. Wired into `propose_questions()`, so it shows on `patient_card()` and `current_questions()` too.
- `app/reasoning/visits/brief.py`: `feeling_note_lines()` lists the note's lines beside the symptom log (its own group in the "changed" section, same fold budget) and records `feeling_note_ids` in `Brief.sources` for provenance. Deliberately **not** duplicated into the brief's own questions preview (which already carries every proposed question) — the same sentence would otherwise print twice on a one-page brief.
[...]

PR #244, opened 2026-09-17
RE-06: the broker v1 and RuleRanker

ashfrnndz21:
## What this is

Story RE-06 (`docs/recommendation-engine.md`, §2.4, build plan row): the broker v1
(`app.delivery.recommend.broker.slate`), the `Ranker` port and its default `RuleRanker`
adapter, and the first four recommendation rules — `new_medicine_explainer`,
`feeling_after_new_medicine`, `visit_topic_week`, `test_coming` — with a conformance
suite for the `Ranker` port.

## Two dependencies aren't actually on `main` yet

- **RE-01** (`Evidence`/`Candidate` types, `FeedItem.private_to`) is still open PR #233
  (62 files, well beyond RE-01's own scope — audit, consent, migrations). Rather than
  depend on an unmerged, oversized PR, `app/delivery/recommend/models.py` vendors the
  exact `Evidence`/`Candidate`/`NoEvidence`/`readable_by` contract from that PR verbatim
  (same fields, same behaviour) — it should collapse into one definition trivially when
  #233 merges. `FeedItem.private_to` is not needed: none of this story's four rules rest
  on search history.
- **RE-05** (PR #243, "What Nura uses" switches) is also unmerged, as expected — the
  brief said to code against its fact shape without merging it. `broker.py` reads
  `subject="signals", attribute=<family>` facts directly (the exact shape #243 defines)
  rather than importing its module.

## Acceptance, line by line

> golden slates match for a fixture profile

`tests/test_recommend_broker.py::test_golden_slate_matches_the_fixture_profile` — a new
medicine (amlodipine), three rising blood-pressure readings, a cloud tap on its
watch-out word, a doctor visit and a lab visit inside the week. All four rules fire; the
same profile ranks the same way on a second read, and re-ranking the same candidates in
reverse input order lands on the same output (determinism + order-independence, not
just repeatability of one code path).

> every candidate's evidence resolves and is `readable_by(context)`

Asserted directly on every candidate in the golden test. `Candidate.__post_init__`
refuses an empty `because` at construction (`NoEvidence`) — never discovered later by a
filter — and `broker.slate()` drops anything `readable_by` refuses as a final,
defensive pass even though every candidate is already built only from what the key
successfully read.

> a switched-off family produces no candidate from it

`test_a_switched_off_family_produces_no_candidate_from_it`. None of this story's four
rules reads a family RE-05 defines (food/sleep/steps/water are RE-10's lifestyle logs,
not built yet), so the mechanism is exercised with a fixture rule declaring
`signal_family="food"` — real production code path (`RecommendationRule.signal_family`,
`broker.slate`'s filtering), not a test-only stub. Writing `signals`/`food`=`False` (the
real fact shape) makes the fixture rule's candidate disappear; every other rule keeps
working.

> a key without a scope never sees a candidate resting on it (named as withheld, not
> silent)

`test_a_key_without_records_never_sees_what_rests_on_it_named_as_withheld`. Kit holds
every caregiver scope but RECORDS; `feeling_after_new_medicine` (needs a cloud tap,
RECORDS) never appears in her slate, `new_medicine_explainer` (MEDICINES alone) still
does, and `Scope.RECORDS` is named in `Slate.withheld`.

> Every read under the `KeyContext` via `audited_read`/`scoped_select`

Every reader in `broker.py` (`_active_lines`, `_recent_taps`, `_upcoming_visits`,
`_signal_is_on`) goes through `audited_read` or `current_facts` (which itself calls
`audited_read`), gated by `context.allows(scope)` first — the same shape
`app.reasoning.patterns.series` already uses, so a scope the key does not hold is left
out and named, never guessed at with a silent empty read.

> Deterministic ordering with an explicit tie-break

`rank.py`'s `_tie_break`: `(rule_id, topic, output, sorted evidence ids)`. Covered by
the golden test and by `test_ranker_conformance.py`'s `tied` case in both adapters.

> No new columns, no migration

None added.

## Scope decisions (read the module docstrings in `rules.py` for the full reasoning)

- `visit_topic_week`'s "a trend on that visit's provider kind" is read as a direction
  over the reading series usually seen for that provider kind
  (`PROVIDER_KIND_READINGS`) — the design's own "kidney test" example names a topic the
  RE-04 catalogue does not carry yet (tests are a later story's topics per `topics.py`'s
  own module doc). This fires for blood pressure and blood sugar, the reading kinds
  that do have a catalogue topic.
- `test_coming` reads "a paper or a visit names a test" as an upcoming visit at a LAB
  provider, tagged by the existing `TopicTagger` against the appointment's own
  `purpose`. The REMINDER-if-fasting half is **not built**: the only fasting fact today
  (`subject="fasting", attribute="today"`) records whether he is fasting *right now*,
  not what a future test will need, and adding that fact would be a new column this
  story does not add. Named as a gap in `rules.py`'s module doc for whichever story
  adds that fact.
- `Ranker.rank`'s `engagement` parameter is part of the port's signature (matches the
  design) but no per-topic engagement counts exist yet — `FeedItem` does not carry a
  topic to count against until RE-07. `RuleRanker` implements the arithmetic
  (`+10`/`−15`) against it defensively; it is exercised in the conformance suite with a
  hand-built mapping, not yet wired to real numbers by the broker.

## Gates (all clean)

- `python3 -m pytest tests/test_ranker_conformance.py tests/test_recommend_broker.py -q`
  → **9 passed**
- `make lint` → ruff: all checks passed; mypy: **Success: no issues found in 272 source
  files**
- `make plain-words` → **8111 strings in 33 files, 0 failures** (no new patient-facing
  strings)
- `make language` → **8257 strings, 2781 keys in 31 catalogues, 0 failures**

Per the operator's review policy, this PR touches medicines, records and keys/scopes —
independent review is required before merge, not just automated checks.

- `FeedItem.private_to` / the `asked_about` rule (RE-01/RE-04's search-history half,
  RE-06's row does not need it).
- Wiring the slate into any output (feed, triggers, visit questions) — that is RE-07.
- The `bring_insurance`, `pattern_found`, `logged_for_doctor`, `asked_about` rules from
  §2.4's table — RE-06's row names only the four built here.
- Real per-topic engagement counts feeding `Ranker.rank`'s `engagement` — no story
  before RE-07 gives them a place to live.

## Risks worth knowing about

- `PROVIDER_KIND_READINGS` and `READING_TOPIC` in `rules.py` are my own stand-in for a
  mapping the design assumes exists somewhere; a pharmacist/clinical reviewer should
  check the provider-kind-to-reading-kind associations make clinical sense before this
  reaches anyone real.
- Two upstream stories (RE-01 PR #233, RE-05 PR #243) are still open. When #233 merges,
  `models.py` here should be deleted in favour of its version (they're identical in
  shape); when #243 merges, `SIGNALS_SUBJECT`/`_signal_is_on` in `broker.py` can import
  `app.reasoning.signals` instead of re-reading the fact shape directly.


ashfrnndz21:
Carried in PR #248 (RE-07), which includes every commit here plus the @fixture mark and the merge with main; #248 is the one that merges.

PR #247, opened 2026-09-17
Fix the #142 restore-race test flake (query-string glob never matched)

ashfrnndz21:
## Summary

`web/tests/e2e/day.spec.ts`'s "a tap in the first second after reopening lands where it was aimed, never on Today underneath it" (#142) was firing intermittently in CI. Reproduced locally: **9/20 failures** with
`npx playwright test tests/e2e/day.spec.ts -g "first second" --repeat-each 20` against the code as it currently exists (see "Branch note" below).

## The race

The test controls the restore race by hand: it holds `/api/doors` open with

```ts
await page.route("**/api/doors", async (route) => {
  await gate;
  await route.continue();
});
```

`nura.doors()` is called with a `?language=en` query string (`web/src/api/nura.ts`). A Playwright glob with no wildcard after `"doors"` only matches that literal suffix — never a URL with a query string trailing it — so `"**/api/doors"` **never matched** the real request, and the route handler was never invoked. The "gate" was a silent no-op: the un-intercepted network call resolved in ~5ms on localhost, racing the test's own `mouse.move`/`mouse.down` call. Whichever won that ~5ms race decided whether the restore guard was armed before the screen flipped to Today.

Confirmed with request/response instrumentation (added temporarily, not part of this diff): across repeated runs, the route handler's own log line never printed, while `/api/doors?language=en` requests consistently completed in a few ms, regardless of whether the test's `gate` promise had been released yet.

## The fix

```ts
await page.route("**/api/doors**", async (route) => {
```

The trailing `**` is the idiom this codebase already uses elsewhere for query-bearing routes (see `player.spec.ts`'s `"**/api/profiles/*/artifacts/*/clip**"`). With it, the gate genuinely holds the request, `restoreGuard`'s pointerdown always arms deterministically before the restore can land, and the swallow-the-stray-tap behaviour is exercised the way the test intends.

**No production code changed.** `restoreGuard.ts` / `main.tsx` are correct as written — this was entirely a test-synchronisation bug, not an app bug. The one line touched is `web/tests/e2e/day.spec.ts`.

## Branch note (please read)

The #142 code (`restoreGuard.ts`, its `main.tsx` wiring, and this e2e test) is **not yet on `main`** — it only exists on `origin/wip/offline-and-the-player` (commit `9993878f`), which is ahead of the actual open PR #221 (`offline-and-the-player`) branch. Since I was asked to branch from `origin/main`, and the file this fix touches doesn't exist there, this branch cherry-picks that one commit first so the fix is testable and the diff against PR #221's eventual content is a single line.

**Risk / what I did not decide:** merging this PR as-is would introduce `restoreGuard.ts`/`main.tsx` to `main` ahead of PR #221, which will likely duplicate that code when #221 merges. The operator may prefer instead to just take the one-line `day.spec.ts` fix (`"**/api/doors**"`) and apply it directly on top of `wip/offline-and-the-player` / PR #221, and not merge this PR's copy of `restoreGuard.ts` into `main` separately. I did not merge or push to any other builder's branch, per the worktree rules.

## Verification

- `npx playwright test tests/e2e/day.spec.ts -g "first second" --repeat-each 20` — **20/20 passed** (was 9 failures/20 before the fix)
- `npx playwright test tests/e2e/day.spec.ts` (full file, once) — **16/16 passed**
- `npx tsc --noEmit` — clean
- `npx vitest run tests/unit/restoreGuard.test.ts` (the only unit file touched, via the cherry-pick) — **4/4 passed**

Dev server run on port 9301 (`NURA_BASE_URL`/`WEB_BASE_URL` overrides), stopped by Playwright's own webServer teardown (its own PID) after each run.


ashfrnndz21:
Taken as the one-line test fix onto PR #221 (the branch that carries restoreGuard), so the cherry-pick isn't needed.

PR #287, opened 2026-09-18
llm: read a model's answer by its text block, never by position

ashfrnndz21:
**Found live by the owner**: uploading a real PDF in onboarding answered "Nura could not read this page" although Claude had read it (API 200). The answer opened with a `thinking` block and `ClaudeExtractor` read `message.content[0].text` → `AttributeError: 'ThinkingBlock' object has no attribute 'text'`, swallowed as a schema mismatch.

- `app/llm/blocks.py` `answer_text(message)`: the text blocks by type, joined in order; `NoTextInAnswer(ValueError)` when there is none.
- Used in the three adapters that read by position: the extractor, the care-navigation drafter, the narrator. The other Claude adapters already walk the blocks.
- Tests: the helper (thinking first, several text blocks among tool blocks, none at all) and the extractor with a thinking block ahead of its answer.

Gates run locally: the three adapters' tests + the new ones (77 passed), `make lint` clean.


ashfrnndz21:
Landed through #288, which contains this commit and passed every check including Postgres.

PR #292, opened 2026-09-21
ingestion: controlled vocabulary + range belongs to its result (checkpoint 2, DATA)

ashfrnndz21:
## Summary
Checkpoint 2 ("Load a paper") — the DATA side. Fixes the three defects the owner hit on a real cholesterol report:

1. **"Another line on the paper" everywhere** — `fieldLabel()` now falls back to the paper's own `label_on_paper` (always given by the extractor when `attribute` is `"other"`, i.e. genuinely outside the vocabulary) before the generic line name, and the extraction prompt/schema teach the model a controlled subject/attribute vocabulary for lab reports: `lipid_panel` (existing) plus new `blood_test` (hba1c, haemoglobin, tsh, uric_acid, white_cells, platelets), `kidney_panel` (creatinine, egfr, potassium, urea, sodium — creatinine/egfr/potassium/hba1c/haemoglobin/tsh match the codes `app.reasoning.ranges`/`app.delivery.trend_strings` already use, for consistency across trend and review screens) and `liver_panel` (alt, ast, alp, ggt, bilirubin, albumin), alongside the existing header codes (`lab_report.patient_name/patient_id/ordering_doctor/facility`). No existing key renamed.
2. **Some lines showed no value at all** — `valueText()` now renders every shape the API can send (number, string, boolean, `{instruction}`, arrays, nested structures) recursively instead of silently dropping anything it doesn't recognise; `readableValueText()` is the last-resort "Nura could not read this line." when there is truly nothing to render, wired into the non-correctable field's own value line in `Records.tsx`.
3. **A reference range arrived as its own unlabelled line** — a result field may now carry `range: {low, high, text}` directly (schema on `ClaudeExtractor`'s `_FIELD_SCHEMA`, no `minimum`/`maximum`, every property typed, `output_config={"format": {"type": "json_schema", ...}}`, read via `app.llm.blocks.answer_text` — unchanged). A server-side normaliser, `fold_legacy_reference_ranges`, folds an older `<analyte>_reference_range` sibling field onto its result (and drops the sibling) so the existing `lab-report-vitals` fixture and any answer in the old shape keep working without a rewrite — a result that already carries its own range is never overwritten.

## What changed
**Backend**
- `app/ingestion/extract.py`: `PrintedRange` + `parse_printed_range()` (table-driven bound parsing: `<150`, `>1.0`, `3.9 - 6.0`/`3.9–6.0`/`3.9 to 6.0`, `Up to 40`, `Negative`, `""`, with unit suffixes); `fold_legacy_reference_ranges()`; `ExtractedField` gains `range`/`label_on_paper`.
- `app/ingestion/claude_extract.py`: `_FIELD_SCHEMA` gains `range` (nested object schema, every property typed, no bounds) and `label_on_paper`, both optional; `_field_from` parses them.
- `app/ingestion/review.py`: `card_from` runs `fold_legacy_reference_ranges` before writing `ReviewField` rows.
- `app/ingestion/models.py` / `app/channels/api/schemas.py`: `ReviewField`/`ReviewFieldOut` gain `range` (JSON) and `label_on_paper` (String(120)), additive.
- `backend/migrations/versions/0052_review_field_printed_range.py`: single linear head, `test_migration.py` pin updated. `scripts/data_map.py` classified both columns as `health`; `docs/trust/pdpa-data-map.md` regenerated.
- `backend/app/llm/prompts/extract_document.txt`: rewritten with the controlled vocabulary, the `range`-belongs-to-its-result rule, and the `other`/`label_on_paper` instruction.
- New fixture `metabolic-panel-2026-09-18` exercises the new vocabulary + the modern `range` shape + the `other`/`label_on_paper` escape hatch end to end through `FixtureExtractor` → `card_from`.

**Web**
- `src/api/types.ts`: `FieldRange`, `ReviewFieldOut.range`/`.label_on_paper`.
- `src/onboarding/review.ts`: `fieldLabel` fallback order (canonical → `label_on_paper` → generic); `valueText` never drops readable content; `readableValueText`; `rangeStatus(value, range) -> "above"|"below"|"in"|"unknown"` (mirrors `app.reasoning.ranges.Range.band_of`'s inclusive/exclusive rules) exported for the future table UI — not wired into any screen (that's the next builder's job, per the brief: I did not restyle or build table UI).
- `src/strings/{en,ms,zh}.ts`: labels for every new canonical key (patient plain words, matching `docs/plain-words.md` §6's registered terms for "kidney number"/"kidney filter"/"body salt"/"sugar test" where they apply) + `records.valueUnreadable`.
- `src/screens/onboarding/Records.tsx`: the non-correctable field's value line uses `readableValueText`.

## How legacy outputs are folded
`fold_legacy_reference_ranges` runs once, in `card_from`, on every extraction regardless of which extractor produced it (fixture or Claude). It scans for `<subject>/<attribute>_reference_range` siblings; when the base result exists and doesn't already carry its own `range`, the sibling's text is parsed into the base's `range` and the sibling is dropped from the card entirely (never shown as a line of its own again). A sibling with no result to fold onto, or whose value isn't text, or whose result already has its own range, is left exactly as it was — nothing is silently discarded.

## Tests
- Backend: **full suite, 2956 passed / 3 skipped**, 0 failed. `make lint` (ruff + mypy) clean, `make plain-words` 0 failures, `make language` 0 failures, `python3 -m scripts.data_map --check` clean.
- New: `tests/test_printed_ranges.py` (22 tests, table-driven range parsing + folding), `test_claude_extractor.py` additions (schema shape, range/label_on_paper round trip), `test_paper_kinds.py` addition (the new metabolic-panel fixture end to end; the existing `ldl_reference_range` test updated to assert the fold instead of a separate field/fact).
- Paper-accuracy harness: still **100.0% safe accuracy**, 18 papers, 92 labelled fields.
- Web: `npm run typecheck` clean, `npx vitest run` **565 passed** (new: never-blank `valueText`/`readableValueText` coverage including booleans/arrays/nested objects, `fieldLabel` fallback ordering, `rangeStatus` for every branch including unparseable ranges), `npm run plain-words` 0 failures, `npm run build` clean.
- E2E: **24/24 passed** on a private port (8012) — `onboarding.spec.ts`, `papers.spec.ts`, `record-papers.spec.ts`, `warm-home.spec.ts`, including a new assertion on the lipid-panel review card that no field is titled with the generic fallback and no field's value is blank.

## What I did not do
- No table UI (`rangeStatus` is exported and unit-tested but not wired anywhere — that's a later builder's job per the brief).
- No visual/styling changes to any screen.
- Did not add ALT/AST/ALP/GGT/bilirubin/albumin to `app.reasoning.ranges`' guideline table or `app.delivery.trend_strings` — those are the liver-panel trend/range-lookup features, out of scope for this DATA checkpoint (the review card shows them read, with no range bar judgement, which is correct — the brief is explicit that `rangeStatus` compares only against the paper's own printed range, never a guideline).
- Left the now-unused `lipid_panel.ldl_reference_range` label key in the strings files rather than removing it, per "do not rename existing keys" (removing an unused key felt like scope creep against that rule, and it is harmless).

Merged `origin/redesign` into this branch (docs-only commits, clean merge) and re-ran `make lint` on the merged head per the brief.

Not merged, per the brief's rule that anything touching extraction gets independent review.


ashfrnndz21:
Addressed the safety review's required changes:

**REQUIRED 1** — \`parse_printed_range\` now requires a whole-string match (only trailing whitespace + a digit-free unit token allowed) and refuses \`low > high\`. Confirmed rejecting: \`"13.0-17.0 (M) / 12.0-15.0 (F)"\`, \`"0.5-0.1"\`, \`"6.0 - 3.9"\`, \`"M: 13-17 F: 12-15"\`, \`"<0.5 (low risk) / 0.5-1.0 (moderate)"\`, \`"5.2 (desirable) / 6.2 (high)"\`, \`"4.0-10.0 x10^9/L"\`, \`"3,9 - 6,0"\`, \`"1,000 - 2,000"\` — all -> null bounds, text kept. Added ≤/≥ alongside </>.

**REQUIRED 2** — \`fold_legacy_reference_ranges\` now sets the merged field's confidence to \`min(base.confidence, legacy.confidence)\`; \`needs_confirm\` is derived from confidence, so a faint range now correctly carries through.

Also done: prompt now explicitly calls out compound/sex-specific/multi-threshold ranges as ambiguous with an example; \`rangeStatus\` has a comment recording the same-row/same-unit invariant; table-driven tests added for every case above plus \`≤/≥\`, a valid negative range, and a parser/value-unit decoupling test; \`rangeStatus\` tests added for a non-numeric string value and exact-bound values per range shape.

**Results:** backend full suite 2974 passed / 3 skipped / 0 failed (was 2956/3). \`make lint\`, \`make plain-words\`, \`make language\`, \`data_map --check\` clean. Web: typecheck clean, vitest 566 passed (was 565 — one existing test needed a one-line exclusion for the \`other\` attribute, which has no canonical label by design), plain-words clean, build clean. E2E on port 8012: \`onboarding.spec.ts\` + \`papers.spec.ts\`, 11/11 passed.

Still not merged.

PR #303, opened 2026-09-21
redesign: What it means for you — the insight screen after Looks right (package 7)

ashfrnndz21:
## Summary
- After a paper is confirmed ("Looks right") — from both onboarding (Records/Plan) and the Record's own Papers list — the person now lands on a real "What it means for you" screen (checkpoint 3, package 7): the backend's own real stages, then the headline, then what stands out on the paper (reusing the report table's own `reportRow`/`rangeStatus`, never a second range-comparison), then the questions for the doctor, then "Keep these questions" (`POST …/insight/keep`), which reports where they landed.
- Onboarding continues afterward through its own "Next" (`continueAfterPaper`, deferred from what `ReviewStep.looksRight()` used to run at once); the Record's Papers flow returns to the list the same way it always did, with the same "Nura wrote it down." note.
- New API client functions (`paperInsightStream`, `keepPaperInsight`) and types, following the existing `insightsStream` pattern; a small additive `signal?: AbortSignal` on `Call`/`sendStream` so leaving the screen aborts the stream — no state update after unmount.
- Strings added in en/ms/zh under a new `paperInsight` namespace, each `// @patient` tagged with self/`…Other` twins.

## Deviations from the literal package-7 spec (with reasons)
- **Selectable question rows are visual only.** `POST …/insight/keep` (already merged, unchanged here) takes no request body and always files every question from the paper's saved insight — confirmed by its own test (`kept_count == len(report["questions"])`, no body sent). The checkboxes let a person mark which ones they still want, and "Keep these questions" is disabled with none checked, but unchecking one does not remove it from what's actually filed (the backend has no way to receive that). Noted as a real gap between the brief and the backend's own already-shipped contract, not something to patch around on the web side.
- **The "where it went" line never names a doctor or a date.** `PaperInsightKeepOut` only carries `appointment_id`, not the visit's own date/doctor; naming those would need an extra `GET /appointments` read this PR intentionally leaves out to avoid a second failure mode for a cosmetic gain. Says "Kept for your next visit." instead.
- **"What stands out" is computed client-side** from the just-confirmed `ReviewCardOut`'s own fields (the same `reportRow`/`rangeStatus` the report table already uses), not from the paper-scoped insight's own response, which carries no raw values or ranges at all — only text.

## A real correctness fix found and made
`ReviewStep.looksRight()`'s `onDone`/onboarding paths discarded `confirmReviewCard`'s own return value and kept using the stale, pre-confirm card. The insight screen (and any future caller) now gets the card `confirmReviewCard` actually returns, with corrections merged.

## Test plan
- [x] `cd web && npm run typecheck`
- [x] `cd web && npx vitest run` — 69 files, 683 tests
- [x] `cd web && npm run plain-words` — 0 failures
- [x] `cd web && npm run build`
- [x] `make plain-words` (root) — 0 failures
- [x] `make language` (root) — 0 failures (found and fixed a real Malay/Chinese wording-consistency failure my own new safety line introduced)
- [x] `make lint` on the merged head (origin/redesign merged in cleanly, docs-only)
- [x] E2E, private port, one worker: `cp3Insight.spec.ts` (new, 6/6), `a11y.spec.ts` (12/12), `onboarding.spec.ts`, `record-papers.spec.ts`, `papers.spec.ts`, `about-him.spec.ts`, `frame.spec.ts`, `warm-home.spec.ts` (35/35) — 53/53 total, all green
- [x] Visual captures (`cp3InsightShots.spec.ts`, not a gate) at 390x844 and 1280x900: mid-thinking, mid-headline, assembled, kept, nothing-stands-out, refusal — all 12 viewed against the blueprint


ashfrnndz21:
## Rework addressed — web rebuild, backend question bank rewrite, two follow-up corrections

### Web (screen rebuild)
- Orb stays beside the headline once the stream finishes (never disappears).
- Headline via SoftText `*word*` markup: `Here is what I would *ask.*` / `Here is what I would ask about {patient}'s *paper.*`
- "Looked at" — quiet chips from the stream's own real data only.
- One Glass card: `For {doctor} on {date}` (real next-visit doctor/date via `nura.appointments()`, same pattern `tabs.tsx`'s `VisitList` uses — no new backend route) or `For your next visit` / caregiver twin. Questions as plain paragraphs inside it.
- Checkboxes and the "What stands out" block removed entirely.
- Buttons stacked full-width: `Keep these for my visit` → `Keeping…` → `Kept`, ConnectionRow line under it; `Not now` (routes wherever "Looks right" used to).
- Safety line once: `Questions to ask, never answers. This is not a doctor's advice.`

### Backend question bank — final EN templates (verbatim)
- `Why is my {label} above the range on this paper?`
- `Why is my {label} below the range on this paper?`
- `Why are {n} of my numbers outside the range on this paper?`
- `Is my {medicine} still the right one for me?`
- `Does this test need to be repeated, and when?`
- THEIRS twins keep the patient's name: e.g. `Why is {patient}'s {label} above the range on this paper?`, `Is {patient}'s {medicine} still the right one for {patient}?`

### Two corrections from the independent read
1. **`patient_card` re-verification gap** — the first pass's two-sentence questions (statement + "What does that mean for me?") passed the pipeline's own gate under a new "ask" kind, but `app.reasoning.visits.questions.patient_card` re-verifies every kept line again under its own default `line` profile (one idea, ≤15 words) — a real `NotPlainEnough` on the actual pre-visit card. Fixed by rewriting every question as genuinely single-sentence; the "ask"-kind plumbing (`plain_words.py` tag regexes, `pipeline.Candidate.verify_kind`) is now unused and reverted. Locked in with a new assertion in `test_keep_is_idempotent_and_lands_on_the_next_visit` that calls `patient_card` after keeping.
2. **Doubled determiner** — `words.ANALYTE_PLAIN_LABEL` and `PLAIN_NAME` now carry the real catalogue's own determiner byte-for-byte ("the bad cholesterol", "your body salt", "ujian gula anda", "您的血糖检查") rather than a hand-bared duplicate. `_bare_word` (renamed from `_bare_medicine_word`, now shared by both the medicine name and the analyte label) strips it before a template supplies its own "my"/"{patient}'s". New parametrized tests (`test_every_value_question_with_every_real_label_never_doubles_its_determiner`, `test_every_medicine_question_with_every_real_medicine_name_never_doubles_its_determiner`, `test_the_aggregate_and_retest_questions_are_clean_in_every_language_and_voice`) render every real label/medicine name × both voices × en/ms/zh through both the plain-words `line` profile and the conclusion-and-advice blocklist. One real blocklist hit surfaced along the way: zh "below" originally used 低于, and 低 is on the zh conclusion-word list (bare-substring match) — never loosened the rule, reworded to 在…范围之下 instead.

### Gates (all green)
- Backend: `ruff check`, `mypy`, `python3 -m app.safety.plain_words` (0 failures), `python3 -m app.language.memory` (0 failures), `pytest tests/test_paper_insight.py` (30 passed), `pytest tests/test_row_scope.py` (11 passed), full suite `pytest -q` (3045 passed, 3 skipped, 0 failed), `python3 -m scripts.data_map --check` (clean).
- Web: `npm run typecheck`, `npx vitest run` (683 passed), `npm run plain-words` (0 failures), `npm run build`.
- E2E (private port, one worker, fresh server, uptime < 10): `cp3Insight.spec.ts` 7/7, plus the full regression list (`record-papers`, `onboarding`, `papers`, `about-him`, `a11y`, `frame`, `warm-home`, `feed`) 56/57 — the one failure (`a11y.spec.ts` "Tab goes through…", a Home-screen tab-order check unrelated to this screen) is a pre-existing flake, confirmed by an isolated rerun passing cleanly; it never touches the insight screen or anything this PR changed.
- Visual: 10 captures at 1280×900 and 390×844 — mid-thinking, mid-headline, assembled, kept, nothing-to-ask, no-visit, caregiver (Mei), refusal — each viewed against the blueprint's own `insight` scene markup and against `scratchpad/shots/cp4-ask/` for spacing parity.

### Not done
Nothing outstanding from the coordinator's list. The optional `visit: {doctor, scheduled_at}` schema addition was not needed — reused the existing `nura.appointments()` fetch instead, as suggested.

This PR still needs an independent safety review of the question bank before merge.

ashfrnndz21:
## Blockers B1–B3 and S1–S4 addressed (S5/S7/S8/S9/S10 and the NOTES left for a follow-up push, as allowed)

Merged `origin/redesign` twice (before and after #306 landed mid-session); resolved conflicts, preferring the merged/operator version wherever both sides touched the same fix (see below).

### B1 — `finalize` never checked `why_plain`
**(a) The shared gate.** I wrote my own fix first (`_clean()` helper, `why_plain` through both `verified()` and `blocked()`). While working, #306 landed on `origin/redesign` with your own fix to the same function (`why_blocked = bool(candidate.why_plain.strip()) and blocked(candidate.why_plain, language)` — blocklist only, not the full plain-words check) plus its own test in `tests/test_analyst_rule.py::test_a_blocked_why_line_fails_the_candidate_exactly_as_blocked_text_does`. Per your instruction I kept **your merged version** — my own version in `pipeline.py` was discarded at the merge (git resolved it automatically to your side; I diffed to confirm and it needed no manual edit). My own weekly-path regression test stayed (different file, no conflict): `test_analyst_claude.py::test_a_clean_text_with_a_blocked_why_plain_is_never_printed_and_files_a_doctor_question` — verified it fails on the pre-fix `pipeline.py` (git-stashed the fix, ran it, got the exact `AssertionError` proving the old code printed the clean text anyway, then restored).

**(b) The paper path.** New `app.reasoning.analyst.paper.choose_from_templates(original, choices)`: reads only `choice.insight_id` off the model's answer, in order — `.text`/`.why_plain` are never read at all. Duplicated ids keep the first; unknown ids drop; fewer than one valid id left falls back to every question `read_phase` found, in its own order. `app.channels.api.analyst._paper_events` no longer calls the weekly `_rebuild` for the paper route. Also added `sanitized_for_model` (S3, below) so the outbound request itself never carries the rendered sentence. Proof test (`test_paper_insight.py::test_the_claude_path_only_chooses_and_orders_never_rewrites_the_papers_own_words`): a fake client returns poisonous `text`/`why_plain` for a valid id — the wire carries the closed-template `Insight`, byte for byte, and none of the poison.

### B2 — medicine-linked question removed entirely
Deleted, not dormant: `_medicine_candidate`, `ANALYTE_DRUG_CLASS_HINTS`, `PAPER_MEDICINE_QUESTION`/`_THEIRS`/`_WHY`, the `"medicines"` read step (`PaperStepKey`), and the `"medicines"` looked-at chip/label entries. The retest closer and the value/aggregate questions stay, unchanged. **A medicine-linked question returns only once a pharmacist-reviewed analyte→medicine mapping exists** — please treat this PR body note as the record of that decision if it needs to live somewhere more permanent.

### B3 — keep with no upcoming visit
Chose the **honest fallback** (option 2): nothing is stored, `kept_count` is always `0`, `filed` is `"unfiled"`, `appointment_id` is `null`. The old standing-memo path (`write_memo`, the fixed `"ask_from_paper"` line) is removed from the route entirely — it named no real question and the reported count was never true either way. The screen says so plainly instead of claiming a keep that did not happen — new `paperInsight.keepNoVisit`/`keepNoVisitOther` (en/ms/zh, caregiver twin): **"There is no visit booked yet, so Nura has not kept these. Book a visit, then open this paper again."** The Keep button itself never shows "Kept" for this path (it reverts to idle — `ThreeStateButton`'s own real-failure behaviour, not a fake success). The proper fix (persisting the real questions unfiled and moving them onto the next booked visit) did not land safely in this pass; flagging for the follow-up push.

### S1 — closed table only
`_analyte_label` no longer takes or falls back to `label_on_paper`. Test plants an unlisted attribute (`lipid_panel.other`, mirroring the real "Apo-B" fixture) and asserts `None`; a second test proves a labelless flagged value is dropped from the single-value question but still counted in the aggregate.

### S2 — printed-range parsing
`_parse_printed_range` now delegates to `app.ingestion.extract.parse_printed_range` (never a second parser). Fixed live: `"5.5 - 3.5"` now refuses instead of reading `(5.5, 3.5)`; `field_range={"low": true}` now refuses instead of becoming `1.0` (fixed at the `_printed_range` call site too, reusing the already-bool-safe `_number()` instead of a bare `isinstance(x, int | float)`). New tests for both, plus the existing sex-specific/compound-text coverage.

### S3 — no PII to the model
New `sanitized_for_model(insight)`: replaces `text`/`why_plain` with a closed kind code (`value`/`aggregate`/`retest`) derived from the insight id's own prefix, keeping id/kind/evidence untouched (evidence only ever sent the caller's own `.id`, never `.label`, already). New serialisation test drives a real caregiver `read_phase` (so the un-sanitised text genuinely carries "Pa" and "cholesterol"), sanitises it, calls the real `_ask_claude` against a fake client, and asserts the actual request body string carries neither.

### S4 — caregiver voice
(i) `keepForVisitOther` added (en/ms/zh) and wired into `Insight.tsx`. (ii) `build_insight` now takes `reader`, resolved in `_paper_events` before the call (moved earlier in the session, since it used to only be fetched after) — `PAPER_HEADLINE_THEIRS`/`PAPER_NOTHING_LINE_THEIRS` are reachable now; new test proves both the "has questions" and "nothing to ask" headlines in both voices. (iii) `cp3Insight.spec.ts`'s caregiver test now asserts the headline and the Keep button exactly, plus a `\bpa\b` word-boundary check in the sweep (the old bare `toContain("pa")` passed against "paper" too).

### Bonus — S6 (needed to re-capture honestly)
`*...*` accent parsing in `web/src/ui/motion.ts` rewritten: parses the accent span across the **whole line** first, then segments for display (Latin/Malay on spaces; Chinese — no spaces at all — via `Intl.Segmenter`/per-character). The old per-space-split match missed a Chinese line entirely and a multi-word accent (`*Mei Ling.*`). New unit tests (en single/two-word, ms, zh, a stray unmatched asterisk) all fail on the pre-fix code (confirmed by stash-and-rerun). Found *live* while re-capturing: `font-style: italic` on a CJK glyph in this font stack produced a real synthesis artifact ("问" rendered with a stray mark through it) — reproduced in isolation, fixed by not italicising CJK accent runs (the font/size still carry the accent).

## Final EN templates (verbatim, unchanged from the last report)
- `Why is my {label} above the range on this paper?`
- `Why is my {label} below the range on this paper?`
- `Why are {n} of my numbers outside the range on this paper?`
- `Does this test need to be repeated, and when?`
- (No medicine question this release — B2.)

## Gate lines
Backend (targeted, not the full suite — the Mac was under heavy load from multiple builders; every run below went through the shared lock script once it existed):
- `ruff check .` — All checks passed.
- `mypy app/` — Success: no issues found in 344 source files.
- `pytest tests/test_paper_insight.py tests/test_analyst_claude.py tests/test_analyst_rule.py tests/test_row_scope.py` — 68 passed.
- `pytest tests/test_visits.py tests/test_visits_api.py` — 52 passed.
- `python3 -m app.safety.plain_words` — 10359 strings in 35 files, **0 failures**, 177 notes.
- `python3 -m app.language.memory` — 10538 strings, 3544 keys in 33 catalogues, **0 failures**.
- `python3 -m scripts.data_map --check` — exit 0.

Web:
- `tsc --noEmit` — clean.
- `vitest run` — 751 passed (70 files).
- `npm run plain-words` — 4800 strings in 3 files, **0 failures**.
- `npm run build` — succeeded.

E2E (private port, one worker, fresh server, uptime < 10; re-run under the shared lock after the load spike — nothing reported below was captured above load 20):
- `cp3Insight.spec.ts` — 7/7.
- `about-him.spec.ts` — 5/5.
- `onboarding.spec.ts` — 6/6.
- `record-papers.spec.ts` — 8/8.
- `ask-layout.spec.ts` — 2/2.
(29/29 total. Three earlier failures — `about-him`'s Home test, `ask-layout`'s wide test, `cp3Insight`'s Record-flow test — were all load-induced: confirmed flakes by isolated re-run once load dropped, never reported as real failures.)

Visual: 24 captures at 1280×900 and 390×844 (mid-thinking, mid-headline, assembled, kept, nothing-to-ask, no-visit, no-visit-kept-nothing [B3], caregiver (Mei), zh [S6], a two-word PATIENT name in Malay "Ah Kong" [S6 — the multi-word accent case is the MS headline twin, which wraps the patient's own name in `*...*`], refusal) — all viewed. The zh and two-word-name ones were re-captured after the italic-CJK CSS fix; visually confirmed clean (no literal asterisks, no stray marks).

## Not done (left for the follow-up push, as explicitly allowed)
S5 (boundary line says "your week" on a one-paper insight), S7 (`keep_paper_insight_questions` not `@audited`), S8 (`latest_paper_insight` reads under `Scope.PROFILE` with no narrowing), S9 (audit-before-call ordering, refusal wrapping in phase 3), S10 (row-scope matrix substitution + mutation proof), and the three NOTES (abort/mounted guard on `nura.appointments()` in `Insight.tsx`, hiding Keep for keys that can never keep, the long-name word-budget fallback to "here").

PR #304, opened 2026-09-21
redesign: Health and the Health Analyst (package 10)

ashfrnndz21:
## Summary

Rebuilds `Health.tsx` and `Insights.tsx` to the blueprint's `health`/`analyst` scenes (docs/design/experience-blueprint.html).

**Health**
- The week ring shows a calm empty state ("No tablets to count yet this week.") instead of the backend's literal "0 of 0" words when a fresh profile has no active medicines (`ringHasNothingToCount`, new `health.ringEmpty`/`ringEmptyOther` strings). The three `health.spec.ts` geometry tests that exercised the old "0 of 0" ring case now cover the two real figures plus a dedicated empty-state geometry test.
- The Health Analyst card leads the screen, its headline drawn with `SoftText` (word-by-word) when a report exists, its existing invitation when none does.
- Sections assemble with `RevealGroup` once, on first mount — a later data refresh (language switch, background refetch) updates the same mounted nodes in place and never re-animates.
- Every existing `data-testid` is unchanged.

**Health Analyst** (`Insights.tsx`)
- Replaces the accumulating `StepTrace` checklist with the orb beside ONE in-place `StatusLine`, driven by the real `step` events `POST …/insights/stream` reports (`records`, `series`, `medicines`, `ledger`, `coverage` — never invented).
- On settle: the headline (`SoftText`), "what Nura looked at" (the real step labels just streamed, once — never shown for a report loaded without a stream having just run), then the report's sections via `RevealGroup`, then the boundary line.
- A new quiet list of past reports (`GET /profiles/{id}/insights/list`, additive, narrowed and audited the same way the existing report reads are; added to `tests/test_row_scope.py`'s conformance matrix), opening one by `GET …/insights/{report_id}`.
- "Generate now" before a report exists, "Look again" once one is on screen.
- Non-happy states: a stream refusal or a dropped connection shows `Notice`'s existing sentence and keeps the last real stage's line on screen ("what arrived stays"); leaving the screen aborts the stream (`AbortController`, a new optional `signal` on `apiStream`/`insightsStream`, additive) and guards every `dispatch`/`setState` behind a `mounted` ref so nothing updates after unmount.
- A pure `analystStreamReducer` (stages → headline → report; a stray late step or a refusal never regresses a settled screen) and `pastReportsNewestFirst` in `web/src/health/insights.ts`, both unit-tested.

**The #2 defect (404 on every fresh Health load)**
`GET /profiles/{id}/insights` used to answer 404 (`NoReportYet`) whenever nothing had been generated yet — correct as a refusal, but read on every Health screen load including a brand-new profile's first one, so the browser logged a failed request on every such load. Backend fix: the route now answers `200` with a `null` body in that case (`latest_report_or_none`); `GET …/insights/{report_id}` keeps its 404 for an id that genuinely doesn't exist, and `latest_report` (raise-and-all) is kept unchanged for the one existing caller that only ever calls it once a report is known to exist. `test_analyst_api.py` updated to match.

## Gates (all run from a clean worktree, `cp5-health-analyst`)

Web (`cd web`):
- `npm run typecheck` — clean
- `npx vitest run` — **686 passed** (68 files)
- `npm run plain-words` — **0 failures**, 43 notes
- `npm run build` — succeeds

Root:
- `make plain-words` — **0 failures**, 176 notes, 3 exempt as history
- `make language` — **0 failures**, 105 notes

Backend (touched — small, scoped changes to `app/reasoning/analyst/service.py` and `app/channels/api/analyst.py`):
- `ruff check app tests` — clean
- `python3 -m mypy app` — clean
- `python3 -m pytest -q` (full) — **3036 passed, 3 skipped**, 0 failed (25m51s)
- `python3 -m scripts.data_map --check` — clean
- `tests/test_row_scope.py` — 11 passed (new `GET …/insights/list` walked)

E2E (own private port 8014, one worker, after `npm run build`):
- New specs: `health-analyst.spec.ts` (17 tests) — all pass
- `health.spec.ts` (updated for the ring's empty state) — 12 tests pass
- `about-him.spec.ts`, `a11y.spec.ts`, `frame.spec.ts`, `design.spec.ts`, `warm-home.spec.ts`, `offline.spec.ts`, `record-papers.spec.ts`, `today.spec.ts` (every spec that mentions `tab-health`, `insights` or `health-`) — 59 tests, 58 passed, 1 failure (`a11y.spec.ts`'s Tab-order test) that reproduces as a pass in isolation and touches no code this PR changed — a pre-existing flake, not a regression.

Visual capture (not a gate): `web/tests/visual/cp5HealthShots.spec.ts` (`-c playwright.visual.config.ts`), written to the scratchpad's `shots/cp5-health/` at 390x844 and 1280x900 — Health top, Health scrolled, fresh-owner Health, analyst mid-thinking (best-effort under CDP network throttling — the real fixture stream is fast enough that this sometimes lands on the same frame as "assembled"; never a wrong or invented frame, only sometimes not a visually distinct one), analyst mid-headline, analyst assembled, past reports, Mei's Health. Every PNG viewed; two real capture-script bugs found and fixed before reporting: the ring/insights-card async state hadn't settled in two of the captures (added explicit waits), and the past-reports capture needed a scroll to actually show the list.

## Deviations from the blueprint (with reason)

- The blueprint's analyst scene shows one bulk action, "Turn these into visit questions." The existing, already-tested infrastructure here is per-insight ("Ask {who} this" → `keep_paper_insight_questions`/visit-questions flow); there is no backend action behind a bulk button distinct from that. Kept the existing per-insight actions rather than invent a bulk one with nothing real behind it.
- "Not enough data for a report" is not a distinct refusal anywhere in the analyst backend — `RuleAnalyst` always produces a report, with each section that has nothing to say rendering its own existing `sectionEmpty` line ("There is nothing here this week."). Left this as the existing, already-correct behavior rather than inventing a second, redundant empty state.

## Test plan
- [x] Health as Pa with seeded data
- [x] Health as a fresh owner (calm empty state, no "0 of 0", no console error during load)
- [x] Generate a report on demand: real stages → headline → sections
- [x] Open a past report
- [x] Mei's view wording (caregiver voice throughout)
- [x] Geometry at 390x844, 360x640, 1280x900


ashfrnndz21:
Merged into `redesign` via the Stage 1 integration PR #306.

PR #308, opened 2026-09-21
redesign: For you — feeds of the day, a first card about him, and live search that finds things (package 13)

ashfrnndz21:
## Summary

**The privacy boundary of outbound queries** (the core of this PR): the feed's self-search
jobs used to hand `ClaudeSearcher` the bare `job.terms` — a generic drug name and nothing
else (`["amlodipine"]`, `["warfarin"]`). The operator's measured diagnosis of the owner's
live test run found 22 of 30 finished jobs returned nothing at all, no items and no
rejections, and every one of those had this shape; the one job that found something had
searched `["cholesterol"]`, already a plain word. This PR does not change what `job.terms`
means or how it is used anywhere else (dedupe keys, batch/medicine matching, hazard and
season checks are all untouched) — it only changes what is *said to the model*:

- `search._safe_queries` builds the query from the licensed catalogue only: his own word for
  a medicine (`app.medicines.strings.PLAIN_NAME`) plus its reviewed purpose group
  (`Monograph.purpose_id`, e.g. "blood pressure" — the same short word the spec's own
  example already used), plus an intent word for the job's kind ("what it is for", "side
  effects to know", "food", "patient information"). A condition, food, hazard or season term
  is already a closed human word and passes through unchanged. **Nothing from a paper, a
  note, a transcript or anything unconfirmed ever reaches this function** — only the job's
  own already-closed term and the catalogue. A new test (`test_feed_search_safety.py`)
  serialises the *full* request the real adapter would send, for a seeded profile carrying
  realistic PII (name, phone, doctor, facility, dates, lab values), through the real
  `plan_learning_jobs` → `search_and_compress` path, and asserts none of that PII appears in
  any request sent to the fake Anthropic client.
- `ClaudeSearcher._ask` now also passes `allowed_domains` (the job's own usable sources) on
  both `web_search` and `web_fetch` tool declarations — restricting the model at the tool
  itself, in addition to (never instead of) the existing after-the-fact allowlist check
  (`_on_allowlist`, unchanged, still enforced on the tool's own returned URL, never the
  model's restated one). Never `blocked_domains` alongside `allowed_domains` (the API
  rejects that pairing).
- Every job now carries `searched_detail` on its `results` — operational-only counts (queries
  sent, capped at 120 chars each; web_search/web_fetch tool-use counts; candidate/allowlist/
  fetch counts; stop reasons) and, when nothing survives, a closed `empty_because` enum
[...]

PR #312, opened 2026-09-21
redesign: For you — the feed in the blueprint's language (package 13, screens)

ashfrnndz21:
## Summary

Stacks on #308 (package 13's backend half) — this PR is currently shown against `redesign`
and so carries #308's commits too until that one merges; once it does, this PR's diff shrinks
to just the three files below.

The owner rejected the feed as "extremely static." The backend PR's own captures showed why:
a white fade bar cutting across every card's top and bottom, a clip's still with no frame, no
line saying why a card is here, the old global header eating ~140px, and four equal pill
buttons dominating every card. This PR is the web half — brief section B, items 7-9.

**Two real rendering defects found and fixed** (`web/src/ui/feed.css`, `web/src/ui/base.css`):
- The "scrolling shadows" technique on every card's scroll region painted `--white`/`#fbf6f0`
  — cream TEXT colour, retokened for the dusk-glass ground in an earlier pass — as an OPAQUE
  background fill at the top/bottom 28px. The glass and paper tiles here are translucent
  washes over a moving gradient ground, not a flat light page, so there was no single opaque
  colour a "cover" could fade to that actually matched what sat behind it — hence the white
  bar over the kicker and the why-link. Dropped the cover layers; the existing plum-tinted
  radial glow (already on-brand, already translucent) is the sole affordance now.
- Found on the way past: `a.pill.plum { color: var(--white) }` painted an anchor's text the
  same cream as its own background, making it invisible. This hit the feed's own new "Watch
  the whole video…" link and a pre-existing one, `family/Consents.tsx`'s "save-record" —
  removed the stale override.

**`web/src/screens/Feed.tsx`**:
- A compact one-row header (`FeedHeader`, the same pattern `AskHeader` already uses) replaces
  the global menu+wordmark+switcher+bell header — back, "More for you", the menu (`open-me`
  kept). The screen's only `h1`.
- Card anatomy reordered: kicker, title (the redesign's conversation head — 25px/1.17/300 —
  for the pager's first card, compact for the rest), body, ONE "why this is here" line (the
  backend's own `why.plain`, RE-08 — never invented, only its visual treatment changed to
  read as reasoning), the publisher, the way into the Why sheet, the boundary last.
- The clip card frames its still at a proper 16:9 with Play centred over it, names its
  publisher as its own byline, shows duration when the item has one, and its one primary
  action is now a solid button, not a text link.
- The four side actions are a quiet icon row: every `data-testid`, label and >=56px target
  `feed.spec.ts`/`formats.spec.ts` already assert is unchanged — only their visual weight, so
  a card's own primary action (Keep going, See your tablets, …) is what the eye lands on
  first.
- The "day's self-searches" status line keeps its own real signal (`GET .../feed/jobs/status`)
  and gains a small orb beside it (existing `Orb`/`StatusLine` kit components) — never a fake
  delay, never a per-job topic this client's signal cannot back (only `looking` exists today).

**Not done** (disclosed, not silently dropped):
- "Reveal on first load only": the existing `Reveal`/`RevealGroup` kit components showed no
  visible `prefers-reduced-motion` guard in the files this PR touched; wrapping the pager's
  cards in them without dedicated verification time risked a real reduced-motion regression I
  could not confirm was safe in the time available.
- "Nothing yet"/quiet-hours restyle: captured and reviewed: they did not show the defects the
  backend PR's own captures flagged, so left as they already render.
- The status line's exact per-topic wording from the brief's own example: no backend field
  carries a per-job topic on this signal yet (disclosed in #308 too).

## Test plan
- [x] `npm run typecheck && npx vitest run` — 734 tests passed
- [x] `npm run plain-words` — 0 failures (no strings changed)
- [x] `npm run build` — clean
- [x] e2e on a private port (8014), one worker, fresh DB, through the shared suite lock:
      `feed.spec.ts` (15), `formats.spec.ts` (5, the clip-card assertions), `about-him.spec.ts`
      (5), `a11y.spec.ts` (12), `frame.spec.ts` (5), `warm-home.spec.ts` (6) — all green on a
      clean DB and reasonable load.
- [x] One real failure found and fixed along the way: `about-him.spec.ts`'s "no caregiver
      screen says nothing" sweep read only the shell's scroll region, which the compact
      header (correctly) now sits outside of — the screen's title text moved out of that
      region with it. Fixed by keeping a non-heading, screen-reader-only echo of the title
      inside the scroll region (not a second landmark — no heading role), rather than
      widening the sweep's own scope across every screen's header.
- [x] `web/tests/visual/cp5FeedShots.spec.ts` re-captured at 390x844 and 1280x900 (8/8
      passed) and every PNG viewed against the blueprint after each fix — the white bars are
      gone, the header is compact, the clip card is framed and its primary action is legible,
      the why line reads as reasoning, the action row is quiet.
- [ ] Full backend suite: not run — this PR has no backend changes; CI covers it.


ashfrnndz21:
Three defects fixed (from the review of the previous captures), all confirmed real and now re-captured/re-verified.

## 1. Clip still: lozenge → 16:9 frame, icon-only Play, duration badge
`.clip-media` was inheriting `var(--radius-chip)` (a large pill radius) instead of a flat frame radius. Now `border-radius: 16px`, capped `max-height: 28vh`, full card width. Play is now `<Icon name="play">` alone in a round 56px button with `aria-label`, not the word "Play". Duration (when the item has one) is a small bottom-right badge, separate from the byline text.

## 2. Byline cut off / primary action invisible
Root cause: the "Watch the whole video…" link lived inside the scrolling `.feed-body`, after the media, why, source and boundary — a short card could push it (and the byline right above it) off-screen with no visible way to know it was there.
Fix: the primary action moved to `.feed-controls` (never scrolled, structurally guaranteed visible — the same place `Keep going`/`See your tablets` already live). A clip card's own body lines are capped to the first 2 before the media; the rest continue right after it, still inside the scrollable region.

Extended `feed.spec.ts`'s "nothing covers a line" test (as asked) to also visit a clip card, and added checks at Pixel 5 **and** 360x640: the primary action is hit-testable and fully inside the card *without scrolling*, above the action row; the clip's media is 16:9 within 2%, radius <= 20px; the byline is never covered once scrolled to (same guarantee every other line already gets — it does *not* need to be scroll-free, only the primary action does, per the original report).

## 3. ~150px empty band above the card
`.feed-card { justify-content: flex-end }` was bottom-aligning a shorter-than-the-page tile. Changed to `flex-start` — the card now sits under the header with the normal gap; spare height goes below the card. Checked across gate, quiet hours, Mei's view, and the clip card — all top-aligned now.

## Gates (all through the suite lock, private port 8014)
- `npm run typecheck` — clean
- `npx vitest run` — 734 passed
- `npm run plain-words` — 0 failures (no strings touched)
- `npm run build` — clean
- `feed.spec.ts` — 10/10 (including the extended geometry test, both viewports)
- `formats.spec.ts`, `about-him.spec.ts`, `a11y.spec.ts`, `frame.spec.ts`, `warm-home.spec.ts` — 33/33
- `cp5FeedShots.spec.ts` — 15/15 across **390x844, 360x640 (new), and 1280x900** — every PNG viewed after each fix, including a synthetic long-title/long-publisher clip card (test-tooling DOM text override on a real render, same technique the network-hold captures already use) to check wrapping since the fixture data's own title/publisher are both short

Also merged `origin/cp5-feed-fixes` (PR #310) and re-confirmed `origin/redesign` is up to date — `make lint`, the affected backend tests and `feed.spec.ts` all re-run clean on the merged head.

PR #318, opened 2026-09-22
register: the register path as one continuous blueprint experience

ashfrnndz21:
## Summary

A new person's phone → code → papers walk fell into pre-redesign screens partway through, even though welcome/sign-in/who/cloud already looked right on their own. This makes the whole path one continuous blueprint conversation, per `docs/design/experience-blueprint.html` and the owner's own word-cloud reference (`docs/design/onboarding-mock.html`).

- **Who is this for**: the blueprint's actual conversation — a greeting, chips (Me / Someone else), his reply as a bubble ("Me. My name is Tan."), Nura's welcome and the privacy line, one Continue — replacing two plain choose-cards for a fresh phone. The caregiver-density switcher (adding another profile) keeps the plain cards.
- **Before we start** (consent): the orb, the versioned words arriving as `SoftText` lines, one pill — no card, no form chrome. Every legal sentence is untouched.
- **About you**: one continuous conversation transcript for the patient density — every answered turn stays on screen, read back in his own words (a right-aligned reply bubble); the next question reveals under it. No more Next/Go back paging. Caregiver density is unchanged.
- **The cloud**: a picked word's follow-up question now opens inline under the cloud, never the separate `asks` screen. The read-back shows every statement on one screen in both densities, never one line a screen.
- **Add a paper**: the blueprint's three rows (icon, bold title, hint) behind the orb's own headline ("Now show me a paper. A blood test helps the most." — this headline is backend-driven during a real sitting, so `backend/app/onboarding/strings.py` needed the matching update too), with "Set up later" straight to Home. Leads straight into the already-redesigned reading flow (`PaperReading.tsx`, untouched).
- **webkit**: added the `webkit` Playwright project (chromium stays the CI default) and walked the whole register path under it — the existing suites plus a new capture spec — with `pageerror`/`console.error` collection. Nothing reproduces the blank screen the owner saw in Safari under Playwright's webkit engine (see report below).
- **Captures**: `web/tests/visual/cp6RegisterShots.spec.ts` walks a fresh phone from Welcome to the first paper's report table, one screenshot per scene, at 390x844 and 1280x900, chromium and webkit — written into the scratchpad and looked at by hand against the blueprint before this PR.

Every step keeps saving exactly what it saved before (settings, cloud picks and answers, read-back verdicts) — only the presentation and navigation changed. Every new string is tagged `@patient` with Malay/Chinese twins; `plain-words`/`language` are at 0 failures.

## The blank-screen finding

Ran the full register-path e2e coverage (`onboarding.spec.ts`, `cp5-onboarding.spec.ts`, `a11y.spec.ts`, the new capture spec) under `--project=webkit`, with `pageerror`/`console.error` collection built into the new capture spec. The register path completes cleanly with zero console or page errors in both engines. Two pre-existing webkit-only test failures were found and are **not** the blank-screen defect: `page.route()` interception is unreliable against this app under webkit when a service worker is active (two `cp5-onboarding.spec.ts` tests that mock a refusal via `page.route`), and Safari's own default Tab-key semantics differ from Chromium's (a keyboard-order test). Neither produces a blank screen; both are test-infrastructure quirks of automating WebKit, not app defects. If the owner can reproduce the blank frame again, the exact repro steps (which screen, orientation, cold start vs. warm) would help pin it down further — it did not reproduce here despite a thorough walk.

## Gates (numbers)

- `python3 -m app.safety.plain_words`: 0 failures (10686 strings, 181 notes — pre-existing baseline)
- `python3 -m app.language.memory`: 0 failures (10865 strings, 105 notes — pre-existing, all on versioned consent text)
- Backend tests (`test_biography.py`, `test_onboarding_conditions_ask.py`, `test_onboarding_settings.py`, `test_onboarding_tell_me.py`, `test_plain_words.py`): 209 passed
- Web unit tests (`npx vitest run`): 820 passed, 73 files
- Web e2e, affected specs, chromium (`onboarding`, `cp5-onboarding`, `a11y`, `cp3Insight`, `midnight`, `papers`, `push`, `today`, `family`): 67 passed, 1 skipped (pre-existing), 2 flaky — both isolated and re-run 3× each, passing every time (a bubble-cloud geometry check and a keyboard-focus check, neither touched by this change)
- Web e2e, register path, webkit: onboarding + cp5-onboarding suites pass; the new capture spec passes at both sizes with zero console/page errors

## What I could not do

- Did not reproduce the blank screen (see above) — reported with evidence rather than guessed at.
- Left `web/tests/visual/cp5OnboardingShots.spec.ts` mechanically patched (the new `who-continue` step, the retired `asks` stage) but did not re-verify its screenshots by eye — lower priority than the new cp6 spec this task asked for.
- Did not touch the "choose many photos" grid's own screen (`web/src/screens/PaperBatch.tsx`) beyond what it already inherits (the `ReadingProgress` it already used) — it is reachable from "Add a paper" but is a secondary path, not the one-paper flow the blueprint's `firstpaper`/`reading` scenes show.


ashfrnndz21:
## Update: the cloud, and CI's webkit fix

Two rounds of follow-up review on this PR:

**The cloud was rejected** against `shots/blueprint-ref/04.png` — it read as a two-column grid of near-identical pills, with a title-plus-grey-subtitle head and an accumulating acknowledgement line, not the blueprint's floating scatter. Fixed:

- `web/src/onboarding/cloud.ts`: `packCloud()` — a deterministic outward-ring search (the same technique word-cloud layouts use) places every word as a circle sized by its own weight tier, none overlapping. `tapHint()` gives Nura one line that changes in place on every tap ("High blood pressure, noted. I added what often goes with it." / "Nura took that off." on an unpick), replacing the old accumulating line. Both are pure and unit-tested — including the exact adversarial case that first exposed a real collision bug (four same-size circles in a row) and the box-vs-circle distinction the real geometry test (`cp5-onboarding.spec.ts`) actually measures.
- `web/src/screens/onboarding/Cloud.tsx`: the head is now the orb beside one conversation line ("What is part of your health? Tap what fits."), never a title over a subtitle.
- New strings split across two lines each (`plain-words` rule 2, one idea a line) so the *composed* sentence still reads as the one line asked for. `plain-words` and `language` are both back at 0 failures.
- Re-captured `cp6RegisterShots.spec.ts` and looked at every cloud frame against `04.png` by hand: a scattered circle cloud, the line changing in place, the bloomed related words, the inline follow-up question both before and after answering.

**CI's webkit failure**: CI has no WebKit browser installed (`browserType.launch: Executable doesn't exist`, all 250 specs on the earlier run). `web/playwright.config.ts` now only includes the `webkit` project locally (`!process.env.CI`) or when `NURA_E2E_WEBKIT=1` is set on a runner that has installed it — chromium stays CI's only project, no `.github` change.

Gates after both rounds: `plain-words` 0 failures, `language` 0 failures, backend tests 209 passed, web unit tests 830 passed, e2e affected specs 69/70 passed (one `family.spec.ts` test flaked on a `door-key` strict-mode collision, unrelated to this PR's area — clean on the immediately preceding run of the same test, and confirmed by re-running the register-path specs in isolation with no such issue).

ashfrnndz21:
Fixed the word-break and empty-bubble defect.

**Root cause of the word breaks**: `overflow-wrap: anywhere` let the browser split a word at any character once a bubble's fixed tier diameter ran out of room. Switched `.word.bubble` to `overflow-wrap: normal; word-break: keep-all; hyphens: none`, which by itself would just let long words overflow their circle — so `packCloud()` now gives every bubble at least `minDiameterFor(word)`: the diameter that fits the longest *unbreakable* word in its label (multi-word labels wrap at spaces, so only the single longest word has to fit), measured with a conservative per-character estimate (no canvas dependency, stays pure and unit-testable in Node and the browser alike).

**The scale-down bug**: my first pass at "shrink everything back down if the cloud runs long" had a real bug — it computed each word's "floor" as `minDiameterFor` (which already includes the tier minimum), so `slack = tier - floor` was always ≤ 0 and nothing ever actually shrank. Fixed by keeping a genuine tier-independent floor (`wordFitOnly`, no tier folded in) separate from each word's own target size. Now ordinary short words shrink back toward their tier's usual size when the cloud is over budget, while a long word's circle only ever shrinks back to what its own word still needs — never below it.

**The empty bubble**: confirmed via a diagnostic e2e walk that no bubble ever has empty `textContent` in the real DOM — it was a paint/screenshot-timing artifact, not a data bug. Fixed defensively: `.word.bubble` now fades in as one unit via `@starting-style` (circle and text mount and fade together), so a revealed bubble can't be caught mid-paint with a circle but no text yet.

**Verified**:
- `npx tsc --noEmit`: clean
- Unit tests: 835/835 passing (added coverage for `minDiameterFor` and `packCloud`'s floor-respecting behaviour)
- e2e: all 21 specs in `cp5-onboarding.spec.ts` + `onboarding.spec.ts` pass on chromium, including the "never overlaps itself" geometry test at 360×640 / 390×844 / 1280×900
- A one-off diagnostic e2e walk through a real cloud (14 bubbles) confirms zero empty bubbles and zero words wider than their own box
- Re-captured `phone-09/10/11/11b` and `wide-09/10/11/11b` (chromium) and looked at every one by hand: no bubble breaks mid-word anywhere, no empty bubble, the organic scatter and inline follow-up card are unchanged
- `plain_words`/`language.memory` gates: 0 failures (no string changes this round)

No new strings, no CSS/behaviour changes outside `.word.bubble` and `packCloud`'s sizing math.

ashfrnndz21:
Fixed the label-overflow defect: "Kidney number checked by a doctor" and "Stroke in the family" now sit fully inside their own circles.

**Root cause**: `minDiameterFor` only ever checked a label's longest single unbreakable *word* against the circle's width. A label with no one long word but several ordinary ones ("Kidney number checked by a doctor") could still wrap to more lines than a circle sized only for its widest word was tall enough for — painting text above and below the circle's own edge.

**Fix**: every bubble now sizes to its whole wrapped *label*, not just its longest word:
- `wrapLineCount`/`widthForLines` find the narrowest width a label can wrap to and still fit within 3 lines — greedy word-wrap over each word's own estimated width, the same rule `overflow-wrap: normal` applies for real.
- `estimateLabelBox` turns that into a width and a height (`lines × line-height`, matching `.word.bubble`'s own CSS `line-height: 1.1`).
- `wordFitOnly` takes whichever of width or height asks for more room — a circle is round both ways at once.
- This applies uniformly to every word `packCloud` places (top or revealed/related) — no special-casing needed, since `minDiameterFor` is already called the same way for both.

**Verified**:
- `npx tsc --noEmit`: clean
- Unit tests: 839/839 passing. Added a test that runs the **whole real conditions graph** (`backend/app/onboarding/conditions.json`, the exact fixture the e2e suite seeds from) — every name, in every language, at every size a pick could ever boost it to — against the box it draws at the bubble's own real font metrics (`estimateLabelBox`/`TIER_FONT_PX`, both now exported for this).
- Added a diagnostic to `cp6RegisterShots.spec.ts` itself: at every cloud screenshot, every bubble's own text (read via a DOM `Range`, not the button's own box) is checked against its circle's real rendered `getBoundingClientRect` — the actual layout, not `packCloud`'s estimate of it. Both capture runs (390×844 and 1280×900) pass this.
- e2e: `cp5-onboarding.spec.ts` + `onboarding.spec.ts` all pass; the "never overlaps itself" geometry test flaked twice on an unrelated pre-existing race (`evaluateAll` reading the cloud mid-re-render right after the "more/fewer words" click, 0 boxes momentarily — reproduces at random viewport sizes, confirmed independent of this change by re-running clean) and then passed cleanly on a follow-up run at all three sizes.
- Re-captured `phone-10/11/11b` and `wide-10/11/11b` and looked at every one by hand: every label, including the two flagged, now sits fully inside its own (larger where it needs to be) circle.
- `plain_words`/`language.memory` gates: 0 failures (no string changes this round)

ashfrnndz21:
Addressed all three independent-review items, plus the minor Doors.tsx fix.

**1. A11Y REGRESSION (focus ring)** — `.paper-row-label` (Records.tsx's `PaperRow`, both the `<label>` wrapping a hidden file input and the bare `<button>` sharing that class for "choose-many") matched neither `label.pill:focus-within` nor any `:focus-visible` rule. Added `.paper-row-label:focus-within, .paper-row-label:focus-visible` in `onboarding.css` — the same treatment `warm.css`'s `.action-row:focus-within` already gives every other row-shaped control. Extended `a11y.spec.ts` with a new test that Tabs through the register path's own first-paper screen and checks all three rows for a ring.

**2. THE CLOUD RENDERS EMPTY ~33% of runs** — confirmed: `Cloud.tsx`'s conditions fetch is a real fetch, and the bubble container painted childless at `height: 0` while it was in flight, indistinguishable from "nothing to show." Fixed: the container never paints empty now — it shows the orb and a loading line (`"Nura is getting your words ready."`, new `@patient` string in en/ms/zh) at a real minimum height until the graph has actually arrived, with `data-loaded="true"`/`"false"` for a test to wait on. Fixed every e2e/visual call site reading the cloud's bubbles via `evaluateAll` (which doesn't auto-wait the way a locator action does) to wait on `data-loaded="true"` first. Verified by hand: slowed the `**/conditions*` route and screenshotted the loading state — orb + "Nura is getting your words ready." — never an empty box.

**3. WEBKIT gated wrongly** — now strictly opt-in (`NURA_E2E_WEBKIT=1` only, dropped the `!process.env.CI` fallback), verified with `--list` across no-env / `NURA_E2E_WEBKIT=1` / `CI=true`. Investigated the sign-in code-field failure separately and reproduced it (1 in 10 repeat runs): WebKit's own `.fill()` on the phone number field intermittently left the controlled input at its unchanged initial `"+65"` rather than the full generated number, so `validPhone` correctly (and silently, by its own documented design) refused to submit — the button dropped back to idle with no error shown, exactly as it's supposed to for an unfinished number. `SignIn.tsx`'s `CodeScreen` has no async gate at all, so this is structurally not item 2's pattern; it never reproduced across dozens of chromium runs this session, only under WebKit's own fill() timing on this machine. Reads as a WebKit/Playwright driver-level input race, not an app rendering defect — moot either way now that webkit never runs by default.

**Minor**: `Doors.tsx`'s `pendingName` (a one-time-use signal) was only cleared on `agree()`'s success path — a failed submit left the old name sitting in it, so a later re-entry down a path that doesn't go through `who` first would reuse the stale name and hide the field again. Now cleared in the catch branch too. Dropped the dead trailing `?? ""` (`pendingName.value` is a signal seeded with `""`, never nullish).

**Verified**: `tsc --noEmit` clean; 839/839 unit tests; `vite build`; 46/46 chromium e2e across `cp5-onboarding`/`onboarding`/`papers`/`record-papers`/`a11y` specs (including the new ring test); 5/5 visual capture specs; webkit gating confirmed via `--list`; merged `origin/redesign` clean (one unrelated file, `Today.tsx`); `plain_words`/`language.memory` gates 0 failures.

